PR #2, opened 2023-08-19
Infra - Scheduler

TomLonergan03:
First draft on a scheduler to allow delaying tasks to arbitrary times in the future without blocking the main thread.

E.g.
If brakes are set to deploy we may need to wait a second to check if they deployed successfully. Instead of doing:
```c++
engageBrakes();
Thread::sleep(1000);
if (!checkBrakesEngaged()) {
  logger_.log(core::kFatal, "Could not engage brakes");
}
```
which blocks the thread for the full second, we instead do:
```c++
engageBrakes();
scheduler.addTask(1000, [&]{
  if (!checkBrakesEngaged()) {
    logger_.log(core::kFatal, "Could not engage brakes");
    return core::kFailure;
  }
  return core::kSuccess;
});
```
with a main loop that is something like:
```c++
while (true) {
  doMainLogic();
  scheduler.run();
}
```
This will keep running our main logic, and switch to the brake engagement check only when it is ready to be checked.

Originally suggested [here](https://github.com/Hyp-ed/hyped-2022/issues/5)


lib/core/scheduler.hpp
```diff
@@ -0,0 +1,42 @@
...
+   * @param delay the minimum delay in microseconds before the task can called
+   */
+  void addTask(uint32_t delay, std::function<core::Result(void)> task);
+
+ private:
+  core::ILogger &logger_;
+  core::ITimeSource &time_;
+  std::priority_queue<std::uint64_t, std::vector<std::uint64_t>, std::greater<std::uint64_t>>
+    task_queue_;
+  std::unordered_map<std::uint64_t, std::function<core::Result(void)>> task_map_;
```
mifrandir, line 39: I am not a fan of the `std::uint64_t`s. They hide what's going on here. In some cases they are meant to be timestamps whereas in others they are essentially task ids.

lib/core/scheduler.hpp
```diff
@@ -0,0 +1,42 @@
...
+  core::Result run();
+  /**
+   * @brief Adds a task to the queue
+   * @note The task will be called at minimum after the delay specified by the task
+   * @note If two task have the same delay, the task added first will be called first
+   *
+   * @param task a function that returns a Result
+   * @param delay the minimum delay in microseconds before the task can called
+   */
+  void addTask(uint32_t delay, std::function<core::Result(void)> task);
```
mifrandir, line 32: `const`s and all that...

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+void Scheduler::addTask(uint32_t delay, std::function<core::Result(void)> task)
+{
+  const auto execution_timepoint
+    = std::chrono::time_point_cast<std::chrono::microseconds>(time_.now())
+      + std::chrono::microseconds(delay);
+  std::uint64_t execution_timepoint_us = execution_timepoint.time_since_epoch().count();
+  // If the timepoint is already in the map, increment it until it is not
+  while (task_map_.find(execution_timepoint_us) != task_map_.end()) {
+    execution_timepoint_us++;
```
mifrandir, line 41: `++x;`

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+void Scheduler::addTask(uint32_t delay, std::function<core::Result(void)> task)
+{
+  const auto execution_timepoint
+    = std::chrono::time_point_cast<std::chrono::microseconds>(time_.now())
+      + std::chrono::microseconds(delay);
+  std::uint64_t execution_timepoint_us = execution_timepoint.time_since_epoch().count();
+  // If the timepoint is already in the map, increment it until it is not
+  while (task_map_.find(execution_timepoint_us) != task_map_.end()) {
```
mifrandir, line 40: why go through all this trouble? 

for one, you could just put something like `struct Task { core::TimePoint scheduled_at; std::function<core::Result(void) handler }` into a priority queue. this is, in my opinion, the correct way of doing this.

secondly, if there is a good reason not to do that, then you should try and group tasks together that have the same scheduled time and have your map have type `std::unordered_map<time, std::vector<handler>>`

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,47 @@
...
+  if (current_time < next_task_time) { return core::Result::kSuccess; }
+  task_queue_.pop();
+  const auto task = task_map_.find(next_task_time);
+  if (task == task_map_.end()) {
+    logger_.log(core::LogLevel::kFatal, "Scheduled task missing callback function");
+    return core::Result::kFailure;
+  }
+  task_map_.erase(next_task_time);
+  const auto result = task->second();
+  return result;
```
mifrandir, line 30: `return task->second();`

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,47 @@
...
+                              .count();
+  const auto next_task_time = task_queue_.top();
+  if (current_time < next_task_time) { return core::Result::kSuccess; }
+  task_queue_.pop();
+  const auto task = task_map_.find(next_task_time);
+  if (task == task_map_.end()) {
+    logger_.log(core::LogLevel::kFatal, "Scheduled task missing callback function");
+    return core::Result::kFailure;
+  }
+  task_map_.erase(next_task_time);
```
mifrandir, line 28: this is expensive.

the immediate optimisation is to do `task_map_.erase(task)` instead, as you have already found the location.

then again, i still think that this is not optimal. try and make use of the priority queue as it's designed to have elements inserted and removed.

lib/core/scheduler.hpp
```diff
@@ -0,0 +1,42 @@
...
+class Scheduler {
+ public:
+  Scheduler(core::ILogger &logger, core::ITimeSource &time);
+  /**
+   * @brief Calls the task at the top of the queue if current time is greater than the delay
+   * specified by the task
+   *
+   * @return kSuccess if no task is called, otherwise returns return value of the task
+   */
+  core::Result run();
```
mifrandir, line 23: I think this should run *all* the tasks that are due.

Further, the return value does not differentiate whether (a) no task was run or (b) a task was run successfully. I believe that this is an important distinction.

mifrandir (changes requested):

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,39 @@
...
+  for (std::size_t i = 0; i < task_queue_.size(); i++) {
+    if (current_time < next_task.sheduled_time) { return result; }
+    core::Result next_result = next_task.handler();
+    if (next_result == core::Result::kFailure) { result = core::Result::kFailure; }
+    task_queue_.pop();
+    if (task_queue_.empty()) { break; }
+    next_task = task_queue_.top();
+  }
+  return result;
+}
```
mifrandir, line 28: The logic flow here seems a little odd to me. I understand why one would write it like this but I am unsure that that's what we want.

I think the logic for something like this should be a simple and clean as possible:
```
while next task is due do
  run next task
  if task failed, return failure
return success
```

I understand that there's a bit more bookkeeping to do. So it would be closer to
```
get current time
get number of tasks scheduled
for at most the number of task scheduled iterations
   get next task
   if task is not due, return success
   run task
   if task failed, return failure
return success
```

Here are some highlights why I think this is better than what you have so far:
1. The call to `.size()` happens before the loop so we avoid any weird behaviour where we take out elements while incrementing `i`, essentially incrementing `i` twice. (i.e. if all tasks are due, then you may only end up running half of them)
2. We removed the extra case of no tasks being due. The extra call to get the current time doesn't seem to be a big deal, and it certainly won't be an issue when the system is under load (because then the queue won't be empty anyways).
3. We avoid the mutable variables `next_task` and `result`
4. The logic has changed a little: I understand a task failing to be an irrecoverable failure on a system level. Maybe we should think about building in some shut down handlers in this case, but I wouldn't expect other tasks to be run after such a failure. Any internal error handling should be done by the tasks through other channels.

lib/core/scheduler.hpp
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * @brief Adds a task to the queue
+   * @note The task will be called at minimum after the delay specified by the task
+   * @note If two task have the same delay, the task added first will be called first
+   *
+   * @param task a function that returns a Result
+   * @param delay the minimum delay before the task can called
+   */
+  void addTask(const core::Duration delay, const std::function<core::Result(void)> handler);
```
mifrandir, line 38: nitpick: for a function called `addTask` I'd expect it to take a task as its main argument.

Just call it `Scheduler::schedule`? Or actually pass in a task.

lib/core/scheduler.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include "types.hpp"
+
+#include <cstdint>
+#include <functional>
+#include <queue>
+#include <unordered_map>
+
+namespace hyped::core {
+
+struct Task {
```
mifrandir, line 14: food for thought: maybe we want to have this live inside the `Scheduler` class, i.e. `core::Scheduler::Task`. I don't feel very strongly about this, but it's worth mentioning because as far as I can tell this struct will only be used inside the `Scheduler`.

test/core/scheduler.cpp
```diff
@@ -0,0 +1,61 @@
...
+  core::Scheduler scheduler(logger, manual_time);
+  bool task_called = false;
+  scheduler.addTask(hyped::core::Duration(0), [&task_called]() {
+    task_called = true;
+    return core::Result::kFailure;
+  });
+  ASSERT_EQ(scheduler.run(), core::Result::kFailure);
+  ASSERT_TRUE(task_called);
+}
+}  // namespace hyped::test
```
mifrandir, line 61: A few tests to add:
1. Lots of tasks, scheduled at the same time.
2. Lots of tasks, scheduled in batches. (I think two batches is enough?)
3. Successful task, failing task, and a later task that never runs (e.g. with an `ASSERT_TRUE(false)` inside)

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+void Scheduler::addTask(const core::Duration delay, std::function<core::Result(void)> handler)
+{
+  core::TimePoint execution_timepoint
+    = std::chrono::time_point_cast<std::chrono::nanoseconds>(time_.now())
+      + std::chrono::nanoseconds(delay);
```
mifrandir, line 34: this seems clunky. maybe you want to add this sort of functionality inside of `core/time.hpp`?

maybe by adding a `+` operator between `TimePoint` and `Duration`?

mifrandir (changes requested):

lib/core/scheduler.cpp
TomLonergan03: How does this work? Seeing as both are defined with `using`. I might be being dumb, but I can't see a way to implement e.g.
```c++
Timepoint TimePoint::operator+(Duration const &rhs) const
{
  return TimePoint(this->time_since_epoch() + rhs);
}
```
as it has to be a member of a class

mifrandir: You might want to write classes that extend/inherit from the underlying types instead of just aliasing with the `using` keyword. Not sure what what the greater implications here would be, but I suspect that we only need a very small subset of the full functionality so it should be fine to write a wrapper.

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,37 @@
...
+Scheduler::Scheduler(core::ILogger &logger, core::ITimeSource &time)
+    : logger_(logger),
+      time_(time),
+      task_queue_()
+{
+}
+
+core::Result Scheduler::run()
+{
+  if (task_queue_.empty()) { return core::Result::kSuccess; }
```
mifrandir, line 14: i'd get rid of this, don't think it adds anything

lib/core/scheduler.cpp
TomLonergan03: While this is a bit clunky, I feel like that might be over the top just to avoid this one instance of verbosity

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+  const auto current_time    = std::chrono::time_point_cast<std::chrono::nanoseconds>(time_.now());
+  const auto number_of_tasks = task_queue_.size();
+  for (std::size_t i = 0; i < number_of_tasks; i++) {
+    const auto next_task = task_queue_.top();
+    if (current_time < next_task.sheduled_time) { return core::Result::kSuccess; }
+    const auto next_result = next_task.handler();
+    task_queue_.pop();
+    if (next_result == core::Result::kFailure) { return core::Result::kFailure; }
+    if (task_queue_.empty()) { break; }
```
mifrandir, line 22: above you return success, here you break?

lib/core/scheduler.cpp
```diff
@@ -0,0 +1,34 @@
...
+  }
+  return core::Result::kSuccess;
+}
+
+void Scheduler::schedule(const core::Duration delay, std::function<core::Result(void)> handler)
+{
+  auto execution_timepoint = time_.now() + std::chrono::nanoseconds(delay);
+  const Task task{execution_timepoint, handler};
```
mifrandir, line 30: We prefer using constructors. Make it easier to add bookkeeping in the future.

mifrandir (approved):
Didn't read the tests but looks fine.

DylanCavers (approved):

PR #4, opened 2023-08-25
Infra - Standardise CMakeLists formatting

TomLonergan03:
Formatting rules are open for discussion

TODO:
- [x] GH action
- [x] .cmake-format should contain all rules, not just changes from defaults


DylanCavers (commented):
this just makes sure all cmake files conform to formatting and we'll manually comply to this? looks good just double-checking in case I'm missing something obvious

TomLonergan03:
> this just makes sure all cmake files conform to formatting and we'll manually comply to this? looks good just double-checking in case I'm missing something obvious

You can use the CMake Format VSCode extension to autoformat, and I have a script also so could add that to `scripts`

DylanCavers (approved):
Looks good, think the inclusion of the script is better to avoid depending on vscode extensions

davidbeechey (approved):

pkhudov (approved):

lsalhani (approved):

licornes-fluos (approved):

PR #5, opened 2023-08-27
Infra - Mqtt Implementation

TomLonergan03:
TODO

- [ ] Connect
- [ ] Publish
- [ ] Subscribe
- [ ] Consume
- [ ] Work out better serialisation system
- [ ] Benchmarks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,8 @@ enable_testing()
 find_package (Eigen3 REQUIRED NO_MODULE)
 include_directories(${EIGEN3_INCLUDE_DIR})
 find_package(Boost REQUIRED)
+find_package(PahoMqttCpp REQUIRED)
+include_directories(${PahoMqttCpp_INCLUDE_DIRS})
 include(FetchContent)
 FetchContent_Declare(
   rapidjson
--- a/lib/core/mqtt.cpp
+++ b/lib/core/mqtt.cpp
@@ -1,7 +1,82 @@
 #include "mqtt.hpp"
 
-#include <mqtt/client.h>
+#include <rapidjson/writer.h>
+
+#include "rapidjson/stringbuffer.h"
 
 namespace hyped::core {
 
+std::optional<std::shared_ptr<Mqtt>> Mqtt::create(ILogger &logger,
+                                                  const std::string &id,
+                                                  const std::string &host,
+                                                  const uint16_t port)
+{
+  std::string address = "tcp://" + host + ":" + std::to_string(port);
+  mqtt::connect_options connection_options;
+  connection_options.set_keep_alive_interval(1);
+  connection_options.set_clean_session(true);
+  auto mqtt_client = std::make_unique<mqtt::client>(address, id);
+  mqtt_client->connect(connection_options);
+  if (!mqtt_client->is_connected()) { return std::nullopt; }
+  return std::make_shared<Mqtt>(logger, std::move(mqtt_client));
+}
+
+Mqtt::Mqtt(ILogger &logger, std::unique_ptr<mqtt::client> client)
+    : logger_(logger),
+      client_(std::move(client))
+{
+}
+
+Mqtt::~Mqtt()
+{
+  client_->disconnect();
+}
+
+core::Result Mqtt::publish(const MqttTopic &topic,
+                           const MqttMessage &message,
+                           const MqttMessageQos qos)
+{
+  rapidjson::Document payload;
+  payload.SetObject();
+  rapidjson::Value header;
+  header.SetObject();
+  header.AddMember("priority", message.header.priority, payload.GetAllocator());
+  header.AddMember("timestamp", message.header.timestamp, payload.GetAllocator());
+  payload.AddMember("header", header, payload.GetAllocator());
+  rapidjson::Value payload_json;
+  payload_json.CopyFrom(message.payload, payload.GetAllocator());
+  payload.AddMember("payload", payload_json, payload.GetAllocator());
+  rapidjson::StringBuffer buffer;
+  auto writer = rapidjson::Writer<rapidjson::StringBuffer>(buffer);
+  payload.Accept(writer);
+  logger_.log(core::LogLevel::kDebug, "Publishing message %s", buffer.GetString());
+  auto pubmsg = mqtt::make_message(mqtt_topic_strings[topic], buffer.GetString());
+  pubmsg->set_qos(qos);
+  client_->publish(pubmsg);
+  return core::Result::kSuccess;
+}
+
+// core::Result Mqtt::consume()
+// {
+//   const auto message = client_->consume_message();
+// }
+
+// MqttCallback::MqttCallback(ILogger &logger) : logger_(logger)
+// {
+// }
+
+// void MqttCallback::connection_lost(const std::string &cause)
+// {
+//   logger_.log(core::LogLevel::kFatal, "Connection lost with MQTT broker: %s", cause.c_str());
+// }
+
+// void MqttCallback::delivery_complete(mqtt::delivery_token_ptr token)
+// {
+//   logger_.log(core::LogLevel::kDebug, "Message delivery complete");
+// }
+
+// void MqttCallback::message_arrived(mqtt::const_message_ptr msg)
+// {
+//   logger_.log(core::LogLevel::kDebug, "Message arrived");
+// }
 }  // namespace hyped::core
--- a/lib/core/mqtt.cpp
+++ b/lib/core/mqtt.cpp
@@ -1,0 +1,7 @@
+#include "mqtt.hpp"
+
+#include <mqtt/client.h>
+
+namespace hyped::core {
+
+}  // namespace hyped::core
--- a/lib/core/mqtt.hpp
+++ b/lib/core/mqtt.hpp
@@ -1,0 +1,30 @@
+#include "logger.hpp"
+#include "mqtt_topics.hpp"
+#include "types.hpp"
+
+#include <mqtt/client.h>
+
+namespace hyped::core {
+class IMqtt {
+ public:
+  virtual core::Result connect()                                                           = 0;
+  virtual core::Result publish(const core::MqttTopic &topic, const mqtt::message &message) = 0;
+  virtual core::Result subscribe(const core::MqttTopic &topic)                             = 0;
+};
+
+class Mqtt : IMqtt {
+ public:
+  Mqtt(ILogger &log, const char *id, const char *host, int port);
+  core::Result connect() override;
+  core::Result publish(const core::MqttTopic &topic, const mqtt::message &message) override;
+  core::Result subscribe(const core::MqttTopic &topic) override;
+
+ private:
+  ILogger &log_;
+  mqtt::client client_;
+  const char *id_;
+  const char *host_;
+  int port_;
+};
+
+}  // namespace hyped::core
--- a/lib/core/mqtt.hpp
+++ b/lib/core/mqtt.hpp
@@ -2,29 +2,84 @@
 #include "mqtt_topics.hpp"
 #include "types.hpp"
 
+#include <optional>
+#include <queue>
+#include <vector>
+
 #include <mqtt/client.h>
+#include <rapidjson/document.h>
 
 namespace hyped::core {
+
+enum MqttMessagePriority {
+  kCritical = 0,
+  kNormal   = 1,
+  kLow      = 2,
+};
+
+enum MqttMessageQos {
+  kAtMostOnce  = 0,
+  kAtLeastOnce = 1,
+  kExactlyOnce = 2,
+};
+
+struct MqttMessage {
+  struct Header {
+    uint64_t timestamp;
+    MqttMessagePriority priority;
+  };
+  Header header;
+  rapidjson::Document &payload;
+  bool operator<(const MqttMessage &other) const
+  {
+    if (header.priority == other.header.priority) {
+      return header.timestamp > other.header.timestamp;
+    }
+    return header.priority < other.header.priority;
+  }
+};
 class IMqtt {
  public:
-  virtual core::Result connect()                                                           = 0;
-  virtual core::Result publish(const core::MqttTopic &topic, const mqtt::message &message) = 0;
-  virtual core::Result subscribe(const core::MqttTopic &topic)                             = 0;
+  virtual core::Result publish(const core::MqttTopic &topic,
+                               const MqttMessage &message,
+                               const MqttMessageQos qos)
+    = 0;
+  // virtual core::Result subscribe(const core::MqttTopic &topic)                             = 0;
+  // virtual core::Result consume()                                                           = 0;
+  // virtual MqttMessage getMessage()                                                         = 0;
 };
 
-class Mqtt : IMqtt {
+class Mqtt : public IMqtt {
  public:
-  Mqtt(ILogger &log, const char *id, const char *host, int port);
-  core::Result connect() override;
-  core::Result publish(const core::MqttTopic &topic, const mqtt::message &message) override;
-  core::Result subscribe(const core::MqttTopic &topic) override;
+  static std::optional<std::shared_ptr<Mqtt>> create(ILogger &logger,
+                                                     const std::string &id,
+                                                     const std::string &host,
+                                                     const uint16_t port);
+  Mqtt(ILogger &logger, std::unique_ptr<mqtt::client> client);
+  ~Mqtt();
+  core::Result publish(const MqttTopic &topic,
+                       const MqttMessage &message,
+                       const MqttMessageQos qos);
+  // core::Result subscribe(const core::MqttTopic &topic);
+  // core::Result consume();
 
  private:
-  ILogger &log_;
-  mqtt::client client_;
-  const char *id_;
-  const char *host_;
-  int port_;
+  ILogger &logger_;
+  std::unique_ptr<mqtt::client> client_;
+  // std::priority_queue<MqttMessage> incoming_message_queue_;
+  bool message_available_;
 };
+
+// class MqttCallback : public mqtt::callback {
+//  public:
+//   MqttCallback(ILogger &logger, std::shared_ptr<Mqtt> mqtt);
+//   void connection_lost(const std::string &cause) override;
+//   void delivery_complete(mqtt::delivery_token_ptr token) override;
+//   void message_arrived(mqtt::const_message_ptr msg) override;
+
+//  private:
+//   ILogger &logger_;
+//   std::shared_ptr<Mqtt> mqtt_;
+// };
 
 }  // namespace hyped::core
--- a/lib/core/mqtt_topics.hpp
+++ b/lib/core/mqtt_topics.hpp
@@ -1,0 +1,10 @@
+#include <string>
+#include <unordered_map>
+
+namespace hyped::core {
+
+enum class MqttTopic { kTest };
+
+static std::unordered_map<MqttTopic, std::string> mqtt_topic_strings = {{MqttTopic::kTest, "test"}};
+
+}  // namespace hyped::core
--- a/src/pod/main.cpp
+++ b/src/pod/main.cpp
@@ -1,6 +1,11 @@
 #include <iostream>
 
+#include <rapidjson/document.h>
+#include <rapidjson/stringbuffer.h>
+#include <rapidjson/writer.h>
+
 #include <core/logger.hpp>
+#include <core/mqtt.hpp>
 #include <core/timer.hpp>
 #include <core/types.hpp>
 #include <core/wall_clock.hpp>
@@ -11,19 +16,26 @@ int main(int argc, char **argv)
   hyped::core::WallClock time;
   hyped::core::Timer timer(time);
   const auto execution_time = timer.measureExecutionTime([time]() {
-    hyped::core::Logger logger("GPIO", hyped::core::LogLevel::kDebug, time);
-    hyped::io::HardwareGpio gpio(logger);
-    auto gpio_reader_opt = gpio.getReader(0, hyped::io::Edge::kNone);
-    if (!gpio_reader_opt) {
-      logger.log(hyped::core::LogLevel::kFatal, "Error");
+    hyped::core::Logger logger("MQTT", hyped::core::LogLevel::kDebug, time);
+    const std::string id   = "test";
+    const uint16_t port    = 8080;
+    const std::string host = "localhost";
+    auto optional_mqtt     = hyped::core::Mqtt::create(logger, id, host, port);
+    if (!optional_mqtt) {
+      std::cout << "Failed to connect to MQTT broker" << std::endl;
       return;
     }
-    auto gpio_reader = *gpio_reader_opt;
-    if (gpio_reader->read() == hyped::core::DigitalSignal::kHigh) {
-      logger.log(hyped::core::LogLevel::kInfo, "High");
-    } else {
-      logger.log(hyped::core::LogLevel::kInfo, "Low");
+    auto mqtt        = *optional_mqtt;
+    const auto topic = hyped::core::MqttTopic::kTest;
+    rapidjson::Document message_payload;
+    message_payload.SetObject();
+    message_payload.AddMember("word", "Hello world!", message_payload.GetAllocator());
+    message_payload.AddMember("number", 42, message_payload.GetAllocator());
+    const auto message = hyped::core::MqttMessage{
+      .header  = {.timestamp = 0, .priority = hyped::core::MqttMessagePriority::kCritical},
+      .payload = message_payload,
     };
+    mqtt->publish(topic, message, hyped::core::MqttMessageQos::kAtLeastOnce);
   });
   std::cout << "Ran for " << execution_time.count() << " ns" << std::endl;
 }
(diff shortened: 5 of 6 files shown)


lib/core/mqtt.hpp
```diff
@@ -0,0 +1,116 @@
...
+};
+
+struct MqttMessage {
+  MqttTopic topic;
+  struct Header {
+    std::uint64_t timestamp;
+    MqttMessagePriority priority;
+  };
+  Header header;
+  rapidjson::Document &payload;
```
DylanCavers, line 35: Problem with the assignment of the payload member of the struct due it being a reference. I just added a constructor of the form
```
MqttMessage(MqttTopic topic, Header& header, rapidjson::Document &payload) : topic(topic), 
    header(header), 
    payload(payload) {};
```
to fix the assignment issue.

src/pod/main.cpp
```diff
@@ -11,19 +16,27 @@ int main(int argc, char **argv)
...
+    auto mqtt        = *optional_mqtt;
+    const auto topic = hyped::core::MqttTopic::kTest;
+    rapidjson::Document message_payload;
+    message_payload.SetObject();
+    message_payload.AddMember("word", "Hello world!", message_payload.GetAllocator());
+    message_payload.AddMember("number", 42, message_payload.GetAllocator());
+    const auto message = hyped::core::MqttMessage{
+      .header  = {.timestamp = 0, .priority = hyped::core::MqttMessagePriority::kCritical},
+      .payload = message_payload,
     };
```
DylanCavers, line 37: Instead of initializing the message like this I used:
```
const hyped::core::MqttMessage message{topic, header, message_payload};
```
although there may be a better way to do this cause manually constructing the header and payload will be annoying every time we send a message.

src/pod/main.cpp
```diff
@@ -11,19 +16,27 @@ int main(int argc, char **argv)
...
+    mqtt->publish(topic, message, hyped::core::MqttMessageQos::kAtLeastOnce);
```
DylanCavers, line 38: No need to have topic as an argument of publish as the topic is already a field of the message, also publish doesn't have topic as an argument

DylanCavers:
After the changes I mentioned above and a few other small ones to do with initialising references in other functions, the only functions that dont compile are `getMessage()` and `consume()`, which I think is to do with the reference member in the struct and how it works with the queue.

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,161 @@
...
+Mqtt::~Mqtt()
+{
+  client_->disconnect();
+}
+
+void Mqtt::publish(const MqttMessage &message, const MqttMessageQos qos)
+{
+  auto constructed_message = messageToMessagePtr(message);
+  logger_.log(
+    core::LogLevel::kDebug, "Publishing message %s", constructed_message->get_payload_str());
```
TomLonergan03, line 40: This should fix the dodgy logs
```suggestion
    core::LogLevel::kDebug, "Publishing message %s", constructed_message->get_payload_str().c_str());
```
`get_payload_str()` returns an `mqtt::string`, which apparently isn't printable (why is there a seperate def of string in paho??)

lib/core/mqtt.cpp
```diff
@@ -99,9 +99,7 @@ mqtt::message_ptr Mqtt::messageToMessagePtr(const MqttMessage &message)
   header.AddMember("priority", message.header.priority, payload.GetAllocator());
   header.AddMember("timestamp", message.header.timestamp, payload.GetAllocator());
   payload.AddMember("header", header, payload.GetAllocator());
-  rapidjson::Value payload_json;
-  payload_json.CopyFrom(message.payload, payload.GetAllocator());
-  payload.AddMember("payload", payload_json, payload.GetAllocator());
+  payload.CopyFrom(*message.payload, payload.GetAllocator());
```
TomLonergan03, line 102: I believe this will overwrite the existing JSON
either call this before adding the header as a member, or by adding another member containing message.payload

lib/core/mqtt.cpp
```diff
@@ -99,9 +99,7 @@ mqtt::message_ptr Mqtt::messageToMessagePtr(const MqttMessage &message)
   header.AddMember("priority", message.header.priority, payload.GetAllocator());
   header.AddMember("timestamp", message.header.timestamp, payload.GetAllocator());
```
TomLonergan03, line 100: also, idk if using `payload.GetAllocator()` with `header` is bad. I did it, but it might cause memory leaks/other weirdness so take a look if this is ok

DylanCavers: Not sure what best practice is here, cause there's no allocator for `rapidjson::Value`'s

TomLonergan03: yeah, I think that was why I did it that way, but I'm pretty sure there should be a correct way to do it

src/pod/main.cpp
```diff
@@ -11,19 +16,26 @@ int main(int argc, char **argv)
...
   hyped::core::Timer timer(time);
   const auto execution_time = timer.measureExecutionTime([time]() {
-    hyped::core::Logger logger("GPIO", hyped::core::LogLevel::kDebug, time);
-    hyped::io::HardwareGpio gpio(logger);
-    auto gpio_reader_opt = gpio.getReader(0, hyped::io::Edge::kNone);
-    if (!gpio_reader_opt) {
-      logger.log(hyped::core::LogLevel::kFatal, "Error");
+    hyped::core::Logger logger("MQTT", hyped::core::LogLevel::kDebug, time);
+    const std::string id     = "test";
+    const std::uint16_t port = 1883;
```
DylanCavers, line 21: change back to 8080

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+core::Result Mqtt::subscribe(const core::MqttTopic topic)
+{
+  const auto topic_string = mqtt_topic_to_string.find(topic);
+  if (topic_string == mqtt_topic_to_string.end()) {
+    logger_.log(core::LogLevel::kFatal, "Attempted to subscribe to nonexistent MQTT topic");
+    return core::Result::kSuccess;
+  }
+  const auto result       = client_->subscribe(topic_string->second);
```
DylanCavers, line 55: segfaults on this line when the callbacks are added

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,164 @@
...
+  return core::Result::kSuccess;
+}
+
+core::Result Mqtt::consume()
+{
+  mqtt::const_message_ptr received_msg;
+  auto received = client_->try_consume_message(&received_msg);
+  if (received) {
+    logger_.log(core::LogLevel::kInfo, "Received message");
+    // auto parsed_message = messagePtrToMessage(received_msg);
```
DylanCavers, line 76: Current problem is here, when we try and parse the received message there is a rapidjson related error

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+namespace hyped::core {
+
+std::optional<std::shared_ptr<Mqtt>> Mqtt::create(ILogger &logger,
+                                                  const std::string &id,
+                                                  const std::string &host,
+                                                  const std::uint16_t port)
+{
+  std::string address = "tcp://" + host + ":" + std::to_string(port);
+  mqtt::connect_options connection_options;
+  connection_options.set_keep_alive_interval(1);
```
TomLonergan03, line 16: What unit is this, and might we want to make it part of the constructor and include in a config file?

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+  connection_options.set_clean_session(true);
+  auto mqtt_client = std::make_unique<mqtt::client>(address, id);
+  auto cb          = std::make_shared<MqttCallback>(logger);
+  mqtt_client->set_callback(*cb);
+  mqtt_client->connect(connection_options);
+  if (!mqtt_client->is_connected()) { return std::nullopt; }
```
TomLonergan03, line 22: log this, `kFatal`

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+  constructed_message->set_qos(qos);
+  client_->publish(constructed_message);
+}
+
+core::Result Mqtt::subscribe(const core::MqttTopic topic)
+{
+  const auto topic_string = mqtt_topic_to_string.find(topic);
+  if (topic_string == mqtt_topic_to_string.end()) {
+    logger_.log(core::LogLevel::kFatal, "Attempted to subscribe to nonexistent MQTT topic");
+    return core::Result::kSuccess;
```
TomLonergan03, line 55: `Result::kFailure` surely lol

TomLonergan03, line 55: also, for a conversation topic, i feel like this is the kind of thing we should catch at compile time with `static_assert`, as I believe we would never have a run time configurable topic? Would like opinions

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+              constructed_message->get_payload_str().c_str());
+  constructed_message->set_qos(qos);
+  client_->publish(constructed_message);
+}
+
+core::Result Mqtt::subscribe(const core::MqttTopic topic)
+{
+  const auto topic_string = mqtt_topic_to_string.find(topic);
+  if (topic_string == mqtt_topic_to_string.end()) {
+    logger_.log(core::LogLevel::kFatal, "Attempted to subscribe to nonexistent MQTT topic");
```
TomLonergan03, line 54: log what the wrong topic was

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+  const auto reason_codes = result.get_reason_codes();
+  for (mqtt::ReasonCode code : reason_codes) {
+    if (code != mqtt::ReasonCode::SUCCESS && code != mqtt::ReasonCode::GRANTED_QOS_0
+        && code != mqtt::ReasonCode::GRANTED_QOS_1 && code != mqtt::ReasonCode::GRANTED_QOS_2) {
+      logger_.log(
+        core::LogLevel::kFatal, "Failed to subscribe to MQTT topic with error code %i", code);
+      return core::Result::kFailure;
+    }
+  }
+  logger_.log(core::LogLevel::kInfo, "Subscribed to topic");
```
TomLonergan03, line 67: log what topic we subscribed to

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+    logger_.log(core::LogLevel::kFatal, "Attempted to subscribe to nonexistent MQTT topic");
+    return core::Result::kSuccess;
+  }
+  const auto result       = client_->subscribe(topic_string->second);
+  const auto reason_codes = result.get_reason_codes();
+  for (mqtt::ReasonCode code : reason_codes) {
+    if (code != mqtt::ReasonCode::SUCCESS && code != mqtt::ReasonCode::GRANTED_QOS_0
+        && code != mqtt::ReasonCode::GRANTED_QOS_1 && code != mqtt::ReasonCode::GRANTED_QOS_2) {
+      logger_.log(
+        core::LogLevel::kFatal, "Failed to subscribe to MQTT topic with error code %i", code);
```
TomLonergan03, line 63: log topic

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+      return core::Result::kFailure;
+    }
+  }
+  logger_.log(core::LogLevel::kInfo, "Subscribed to topic");
+  return core::Result::kSuccess;
+}
+
+core::Result Mqtt::consume()
+{
+  for (std::uint32_t i = 0; i < 100; i++) {
```
TomLonergan03, line 73: can we test how long this takes if it consumes 100 messages, I assume it's very little and it's unlikely we will have more than 100 per cycle of the main loop, but I'd like to be sure.
Can be a TODOLater tbf

lib/core/mqtt.cpp
TomLonergan03: can you check this isn't leaking any memory jic

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+MqttCallback::MqttCallback(ILogger &logger) : logger_(logger)
+{
+}
+
+void MqttCallback::connection_lost(const std::string &cause)
+{
+  if (!cause.empty()) {
+    logger_.log(core::LogLevel::kFatal, "Connection lost with MQTT broker: %s", cause.c_str());
```
TomLonergan03, line 161: is there any way we can relay this back to the main thread? would be good to know when we lose connection, but that will also get caught by the heartbeat so might not be neccessary

TomLonergan03: also, as this is in a different thread I should take a look at logger and figure out if it's threadsafe (it's not (my problem not yours (if you need work for the hackathon maybe???)))

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,174 @@
...
+  }
+}
+
+void MqttCallback::delivery_complete(mqtt::delivery_token_ptr token)
```
TomLonergan03, line 165: what are in these args? i assume it's not relevant but worth checking, same goes for `message_arrived`

lib/core/mqtt.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <optional>
+#include <queue>
+#include <vector>
+
+#include <mqtt/client.h>
+#include <rapidjson/document.h>
+
+namespace hyped::core {
+
+enum MqttMessagePriority {
```
TomLonergan03, line 16:
```c++
// all messages with a priority are processed before any lower priority message is processed
```
or something to that effect

lib/core/mqtt.hpp
```diff
@@ -0,0 +1,121 @@
...
+ public:
+  /**
+   * @brief Sends a message to its topic
+   *
+   * @param message
+   * @param qos
+   * @return core::Result
+   */
+  virtual void publish(const MqttMessage &message, const MqttMessageQos qos) = 0;
+  virtual core::Result subscribe(const core::MqttTopic topic)                = 0;
```
TomLonergan03, line 59: docstring for completeness

lib/core/mqtt.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * @param message
+   * @return MqttMessage if message contains valid header, nullopt otherwise
+   */
+  std::optional<MqttMessage> messagePtrToMessage(std::shared_ptr<const mqtt::message> message);
+
+  ILogger &logger_;
+  std::unique_ptr<mqtt::client> client_;
+  std::priority_queue<MqttMessage> incoming_message_queue_;
+  std::uint32_t messages_in_queue;
+  mqtt::callback_ptr cb;
```
TomLonergan03, line 107: these 2 aren't used anywhere afaict

lib/core/mqtt.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * @brief Sends a message to its topic
+   *
+   * @param message
+   * @param qos
+   * @return core::Result
+   */
+  virtual void publish(const MqttMessage &message, const MqttMessageQos qos) = 0;
+  virtual core::Result subscribe(const core::MqttTopic topic)                = 0;
+  /**
+   * @brief Pulls all pending MQTT messages into internal queue
```
TomLonergan03, line 61: pulls next 100 messages

lib/core/mqtt_topics.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <string>
+#include <unordered_map>
+
+namespace hyped::core {
+
+enum class MqttTopic { kTest };
+
+static std::unordered_map<MqttTopic, std::string> mqtt_topic_to_string
```
TomLonergan03, line 10: `const static` for both of these, maybe `constexpr` if it compiles but i don't think it does

TomLonergan03 (commented):
Looks pretty good

CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ enable_testing()
 find_package (Eigen3 REQUIRED NO_MODULE)
 include_directories(${EIGEN3_INCLUDE_DIR})
 find_package(Boost REQUIRED)
+find_package(PahoMqttCpp REQUIRED)
+include_directories(${PahoMqttCpp_INCLUDE_DIRS})
 include(FetchContent)
 FetchContent_Declare(
```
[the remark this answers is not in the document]
DylanCavers: I'll have a look

lib/core/mqtt.cpp
DylanCavers: that would make sense lmao

DylanCavers: would that work as a `static_assert` inside subscribe? That feels more elegant

DylanCavers: acc on second thoughts is it not impossible for us to subscribe to an undefined topic? As the topic passed in is a member of `core::MqttTopic`, so a topic has to be defined for us to even pass it into the function. I may be thinking about this wrong though

lib/core/mqtt.cpp
DylanCavers: Will try test it soon shouldn't be difficult

lib/core/mqtt.cpp
DylanCavers: This is what they define `delivery_token` as
```
/**
 * Provides a mechanism to track the delivery progress of a message.
 * Used to track the the delivery progress of a message when a publish is
 * executed in a non-blocking manner (run in the background) action.
 */
```
So looks pretty harmless and probably not of much use. The `const_message_ptr` is simply a ptr to the message that has been received, same as we get from the `try_consume_message`. I was originally going to try and use the pointer here to add the messages to the queue but proved difficult.

lib/core/mqtt.hpp
DylanCavers: don't think `messages_in_queue` is used but the `callback_ptr` is there cause otherwise we segfault whenever a callback is called. I think it is cause we move the ptr for the client after we set the callbacks, so we need to move a copy of the callbacks object with it so it can find them when they are called. That's my understanding but may be wrong.

lib/core/mqtt_topics.hpp
DylanCavers: this is the same issue that cropped up [here](https://github.com/Hyp-ed/hyped-2024/pull/23#discussion_r1388694526), do we want them static or const? as having them as both complicates things (define and instantiate in different files)

TomLonergan03: `const` then.
Reading into it, I think a `const` variable at the namespace level is implicitly `static` cos it's obviously only instantiated once. 

(explicitly using the `static` keyword in namespace scope limits it's visibility to the current translation unit only, but what that actually impacts semantically i'm not sure)

lib/core/mqtt.cpp
TomLonergan03: can we `static_assert` that the topic is in `mqtt_topic_to_string`? if we can it would be nice to get a compile time error that we had forgotten to add it to the hashmap

lib/core/mqtt.cpp
TomLonergan03: fair enough

lib/core/mqtt.cpp
DylanCavers: cool will add that

lib/core/mqtt.cpp
DylanCavers: Appears to be longest time no messages pass in seconds.
```
/**
	 * Sets the "keep alive" interval.
	 * This is the maximum time that should pass without communications
	 * between client and server. If no messages pass in this time, the
	 * client will ping the broker.
	 * @param keepAliveInterval The keep alive interval in seconds.
	 */
```
Does 1 second seem like a good amount of time?

lib/core/mqtt.hpp
DylanCavers: `callback_ptr` also needed per paho mqtt maintainer [here](https://github.com/eclipse/paho.mqtt.cpp/issues/141#issuecomment-375402234) lmao

TomLonergan03: can you add a comment saying this is why this is here pls

Will save someone some pain in a years time

lib/core/mqtt.cpp
DylanCavers: "Ran for 1397822 ns"

lib/core/mqtt.cpp
DylanCavers: constant to be added for the keep alive interval

lib/core/mqtt.hpp
```diff
@@ -111,6 +111,7 @@ class Mqtt : public IMqtt {
   ILogger &logger_;
   std::unique_ptr<mqtt::client> client_;
   std::priority_queue<MqttMessage> incoming_message_queue_;
+  // callback_ptr is here to ensure that the callback object is not destroyed
```
TomLonergan03, line 115: add a link to the issue that says this is the approved fix, cos this comment still makes it seem kinda insane lol

Dockerfile
```diff
@@ -3,24 +3,16 @@ ARG DEBIAN_FRONTEND=noninteractive
...
-RUN mkdir build
-WORKDIR /pahocpp/paho.mqtt.cpp/build
-RUN cmake -DPAHO_BUILD_DOCUMENTATION=FALSE -DPAHO_BUILD_SAMPLES=FALSE -DPAHO_MQTT_C_PATH=/pahoc/paho.mqtt.c ..
-RUN make
-RUN make install
+WORKDIR /home
+RUN git clone --recurse-submodules https://github.com/eclipse/paho.mqtt.cpp.git
+WORKDIR /home/paho.mqtt.cpp
+RUN git checkout ef021805c7aca42d72194d85bb2bdccdfe963e28
+RUN cmake -Bbuild -H. -DPAHO_WITH_MQTT_C=ON -DPAHO_BUILD_STATIC=ON -DPAHO_BUILD_DOCUMENTATION=OFF -DPAHO_BUILD_SAMPLES=ON
```
TomLonergan03, line 13: is building samples necessary?

Dockerfile
```diff
@@ -3,24 +3,16 @@ ARG DEBIAN_FRONTEND=noninteractive
...
+RUN sudo cmake --build build/ --target install
+RUN sudo ldconfig
```
TomLonergan03, line 15: what does this do?

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+  logger_.log(core::LogLevel::kInfo, "Subscribed to topic: %s", topic_string->second.c_str());
+  return core::Result::kSuccess;
+}
+
+core::Result Mqtt::consume()
+{
+  for (std::uint32_t i = 0; i < 100; i++) {
+    mqtt::const_message_ptr received_msg;
+    auto received = client_->try_consume_message(&received_msg);
```
TomLonergan03, line 79: `const`?

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,179 @@
...
+    if (!received) {
+      logger_.log(core::LogLevel::kDebug, "Consumed %i message(s)", i);
+      return core::Result::kSuccess;
+    }
+    auto parsed_message = messagePtrToMessage(received_msg);
```
TomLonergan03, line 84: `const`?

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,179 @@
...
+    if (!parsed_message) {
+      logger_.log(core::LogLevel::kFatal, "Failed to parse MQTT message");
+      return core::Result::kFailure;
+    }
+    incoming_message_queue_.push(*parsed_message);
+  }
+  logger_.log(core::LogLevel::kDebug, "Consumed 100 message(s)");
```
TomLonergan03, line 91: "Consumed 100 messages"

Dockerfile
DylanCavers: from what I can tell it's something to do with updating symlinks and keeping track of versions, so idk if its necessary. do you think its worth keeping?

TomLonergan03: where did you get it from?

if it's in the official install guide (and you trust the guide lol) then it should stay

DylanCavers: yeah it was just part of the install guide so probs best to keep it

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include "rapidjson/stringbuffer.h"
+
+namespace hyped::core {
+
+std::optional<std::shared_ptr<Mqtt>> Mqtt::create(ILogger &logger,
+                                                  const std::string &id,
+                                                  const std::string &host,
+                                                  const std::uint16_t port)
+{
+  std::string address = "tcp://" + host + ":" + std::to_string(port);
```
TomLonergan03, line 14: const

lib/core/mqtt.cpp
```diff
@@ -0,0 +1,179 @@
...
+  mqtt::connect_options connection_options;
+  connection_options.set_keep_alive_interval(kKeepAliveInterval);
+  connection_options.set_clean_session(true);
+  auto mqtt_client = std::make_unique<mqtt::client>(address, id);
+  auto cb          = std::make_shared<MqttCallback>(logger);
```
TomLonergan03, line 19: can this be const?

TomLonergan03 (commented):
last 2 nitpicks lol

licornes-fluos (approved):

PR #12, opened 2023-10-15
TEL - Public Facing App Backend

davidbeechey:
Implements the following API routes in the existing NestJS telemetry server for use by the public-facing app:

- `/pods/pod_1/public-data/velocity` - gets velocity measurements (in the given time window)
- `/pods/pod_1/public-data/displacement` - gets velocity measurements (in the given time window)
- `/pods/pod_1/public-data/levitation-height` - gets levitation height (in the given time window)
- `/pods/pod_1/public-data/state` - gets the current state and previous state of the pod
- `/pods/pod_1/public-data/launch-time` - gets the time since the pod changed from `READY` to `ACCELERATING` state

(Time windows are specified using query params `?start=[TIMESTAMP]&end=[TIMESTAMP]`. Start is mandatory, end is optional and will default to now.)

Other features to come at a later date (so we can start implementing the GUI now) are:

- System statuses
- Uptime


Dsavvi180 (approved):
Looks good!

telemetry/packages/constants/src/pods/states.ts
```diff
@@ -29,3 +29,26 @@ export const ALL_POD_STATES = {
...
+  'ACTIVE',
+  'NULL',
+] as const;
+
+export const getStateType = (
+  state: string,
+): (typeof ALL_POD_STATE_TYPES)[number] => {
+  if (FAILURE_STATES[state as keyof typeof FAILURE_STATES]) {
+    return 'FAILURE';
+  } else if (PASSIVE_STATES[state as keyof typeof PASSIVE_STATES]) {
```
arjunnaha, line 45:
```suggestion
  }
  
  if (PASSIVE_STATES[state as keyof typeof PASSIVE_STATES]) {
```

You don't need the `else if`s here because you're returning (i.e. just early return! 😄)

telemetry/packages/server/src/modules/influx/errors/InfluxServiceError.ts
```diff
@@ -0,0 +1,6 @@
+export class InfluxServiceError extends Error {
```
arjunnaha, line 1: ❤️ this

telemetry/packages/server/src/modules/measurement/MeasurementReading.types.ts
```diff
@@ -1,10 +1,39 @@
...
-});
+export const MeasurementReadingSchema = z
+  .object({
+    podId: zodEnumFromObjKeys(pods),
+    measurementKey: z.string(),
+    timestamp: z.string(), // to handle nanoseconds timestamp
+    value: z.number(),
+  })
+  // Validate measurement exists and enum value is valid (if applicable)
+  .refine(
```
arjunnaha, line 13: One thing you could potentially do here as well is validate integers/floats...buttttt that sounds like it might be painful. Could potentially do it 'loose'-ly:

I.e. for integers

`127.1` does not validate
`127.0` does validate
`127` does validate

telemetry/packages/server/src/modules/public-data/PublicData.service.ts
```diff
@@ -0,0 +1,119 @@
+import { HttpException, Injectable, LoggerService } from '@nestjs/common';
+import { Logger } from '../logger/Logger.decorator';
+import { INFLUX_TELEMETRY_BUCKET } from '../core/config';
+import { flux } from '@influxdata/influxdb-client';
+import { InfluxService } from '../influx/Influx.service';
+import { ACTIVE_STATES } from '@hyped/telemetry-constants';
+
+type InfluxRow = {
+  _time: string;
```
arjunnaha, line 9: Do you mind creating a common type `InfluxRow`, shoving it somewhere and then extend from it here? I can see a few other `InfluxRow`s around the codebase

arjunnaha (changes requested):

telemetry/packages/constants/src/pods/states.ts
davidbeechey: Haha quite right - copilot wrote this, should've checked it 😂

telemetry/packages/server/src/modules/measurement/MeasurementReading.types.ts
davidbeechey: Good idea! Can get the format from the pods object. Will add this :)

telemetry/packages/server/src/modules/public-data/PublicData.service.ts
davidbeechey: Will do 👍

arjunnaha (approved):
LGTM!

telemetry/packages/server/src/modules/influx/Influx.service.ts
```diff
@@ -1,43 +1,69 @@
...
-import { Injectable, OnModuleInit } from '@nestjs/common';
+import { Injectable, LoggerService, OnModuleInit } from '@nestjs/common';
 import {
   INFLUX_TELEMETRY_BUCKET,
   INFLUX_ORG,
   INFLUX_TOKEN,
   INFLUX_URL,
   INFLUX_FAULTS_BUCKET,
 } from '../core/config';
+import { Logger } from '../logger/Logger.decorator';
```
TomLonergan03, line 10: Not knowing anything about TS, is it normal to use relative paths?

davidbeechey: Yeah it's pretty normal, but could convert them all to absolute if you want?

TomLonergan03: if this is how you do them idm. I prefer absolute generally, cos it makes it easier to work out where things are but I'll bow to whatever the existing standard is

TomLonergan03 (approved):

PR #17, opened 2023-10-24
TEL - Live server logs on the GUI

davidbeechey:
Live server logs are now shown on the GUI! 

This has been implemented using a custom Winston transporter which automatically sends the logs over a websocket. The GUI listens to the connects to a websocket and displays the logs it receives.

Also some minor changes to the controls.

Still to do:

- [x] Use Tailwind config for theme colours
- [x] Move config values into environment variables
- [x] ~~Think about multiple pod support (might come as a later PR)~~ will deal with this in a separate PR
- [x] ~~Save current tab state in the URL~~ not possible without overcomplicating, could switch to something like Astro or Next.js for the entire frontend if something else justifies this
- [x] ~~Might as well do this at the same time: have separate connection statuses for connection to server, MQTT broker and pod~~ New PR for this!
- [x] Add an input for specifying the levitation height (publish to MQTT topic)
- [x] Update screenshots
- [x] ~~Update controls~~ New PR for this!

<details><summary>Screenshots:</summary>
<p>

image

image

image

</p>
</details>


Dsavvi180 (approved):
All seems well

TomLonergan03 (approved):

PR #18, opened 2023-10-26
MGT - Revamp Debug Binary

kshxtij:
Improvement on basic QoL aspects of the debug binary. 

This PR adds a new dependency of ncurses, can be installed via `libncurses5-dev` on apt and `ncurses` on brew. 

New Additions:
 - Up and Down arrow for history
 - Escape to quit
 - Tab to autocomplete

ToDo: 

- [x] Verify behaviour on windows/wsl
- [x] Finish porting over functionality for shared IO objects/config handling
- [x] Port over commands
   - Filter out commands and classes you don't need for this year
   - Ensure we don't end up in the mess of one 1000 line long file having all the command logic
- [x] Revamp config based on the new TOML configurations
- [x] Possible inclusion of aliasing (preferably within the config because writing `i2c set 0x70 register 0x32` or wtv is a nightmare while in the lab or sitting on the track)


lib/debug/CMakeLists.txt
```diff
@@ -1,15 +1,16 @@
...
   "${CMAKE_SOURCE_DIR}/lib"
   "${CMAKE_CURRENT_SOURCE_DIR}"
   ${rapidjson_SOURCE_DIR}/include
 )
 target_link_libraries(${target} 
+  hyped_core
   hyped_io
   hyped_sensors
   hyped_motors
+  Curses
```
DylanCavers, line 15: Build worked for me by changing `Curses` to `${CURSES_LIBRARIES}`

TomLonergan03: This fixes it for me also

lib/debug/commands/spi_commands.cpp
```diff
@@ -0,0 +1,81 @@
...
+      };
+      auto spi_write_byte_command = std::make_unique<Command>(spi_write_byte_command_name,
+                                                              spi_write_byte_command_description,
+                                                              spi_write_byte_command_handler);
+      repl->addCommand(std::move(spi_write_byte_command));
+    }
+  }
+  return core::Result::kSuccess;
+}
+
```
DylanCavers, line 80: inconsistent newlines at end of file, could linter catch this??

TomLonergan03: very annoyingly, the `clang-format` doesn't have a rule that can catch this. Afaict the standard is newlines before and after namespace declarations, but this isn't 100% consistent. I've fixed it to match in this case, but tbh if the linter can't catch it it's prob not worth the effort to enforce in the long run? Idk how you feel, open to disagreement

DylanCavers: yeah I'd agree, it's nice but I'm not overly bothered by it, probably not worth enforcing if we can't catch it in the linter

lib/debug/commands/gpio_commands.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
DylanCavers, line 1: whitespace below pragma, this might be the same as the whitespace at the end of the files in the sense we can't enforce this, depends what we want to do

lib/debug/repl.cpp
```diff
@@ -1,1028 +1,217 @@
...
-{
-  while (true) {
-    handleCommand();
-  }
-}
-
-std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
+std::optional<std::shared_ptr<Repl>> Repl::create(core::ILogger &logger,
+                                                  Terminal &terminal,
+                                                  const std::string &filename)
```
DylanCavers, line 20: idk if its worth mentioning somewhere exactly what the input file is, but this might be better as a wiki thing

DylanCavers (commented):
Can't comment much on the sensor stuff, but the rest looks good and works for me in wsl

lib/debug/commands/gpio_commands.hpp
kshxtij: @TomLonergan03 If this is something you care about making a Github Action that greps for it is probably doable like some of the ones in lint.yml that I made for `hyped::` and std prefixes for the int types.

lsalhani (approved):

DylanCavers (approved):
lgtm

TomLonergan03 (approved):

PR #19, opened 2023-10-26
MTC - PID controller

lsalhani:
PID controller initial code


TomLonergan03 (changes requested):
This is either C or very close to C. C++ has native class support so we don't create a struct and pass pointers around, instead we define a class in the header file:
```c++
class PidController {
public:
  // public methods here
private:
  // private methods here
  // data members here (i.e what is in the struct now)
}
```
then, in your .cpp file, you define each method in the form:
```c++
PidController::methodName(/*args*/) {
  // do stuff here
}
```
There is more stuff to be changed, but this is the first step to get it into C++, then I will go through again.

(As a side note whatever email you've got set on git isn't the one you're using on github, so either add that email to your github or change the git email, this way commits you make are tied to your account)

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+void PIDController::PIDController_Init()
+{
+  /* Clear controller variables */
+  integrator      = 0.0f;
+  prevError       = 0.0f;
+  differentiator  = 0.0f;
+  prevMeasurement = 0.0f;
+
+  out = 0.0f;
+};
```
TomLonergan03, line 12: When this is refactored to a constructor, it should take args for each of the values we want to configure e.g. `kp`, `ki`, `kd`, maybe limits? (though they may be constexprs somewhere). The data members are then initialised in the initialiser list, so the constructor would look something like this:
```c++
PidController::PidController(const core::Float kp, const core::Fload ki, ...) : kp_(kp), ki_(ki), ...
{
/* any other code that needs run goes here, but most likely this is empty */
}
```

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+float PIDController::PIDController_Update(float setpoint, float measurement)
+{
+  // Error signal
+  float error = setpoint - measurement;
```
TomLonergan03, line 17: `const`, comment is unnecessary

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  // Proportional term
+  float proportional = Kp * error;
```
TomLonergan03, line 20: `const` + don't need comment

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+  prevMeasurement = 0.0f;
+
+  out = 0.0f;
+};
+
+float PIDController::PIDController_Update(float setpoint, float measurement)
+{
+  // Error signal
+  float error = setpoint - measurement;
+
```
TomLonergan03, line 18: lots of unnecessary whitespace throughout, generally avoid it. (if you have a big enough block of code for that to get unpleasant then you should refactor into functions)

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+  // Proportional term
+  float proportional = Kp * error;
+
+  // Integral term
+  integrator = integrator + 0.5f * Ki * T * (error + prevError);
+
+  // Integrator Anti wind-up (dynamic integrator clamping)
```
TomLonergan03, line 25: real nitpicky, but capitalisation is a bit random through a lot of comments. 

btw, this is a good use of comments. generally, we want comments to explain ***why***, whereas the code should be self explanatory as to the ***how*** in most cases (if you do something really weird that is necessary that is worth commenting tho)

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+    integrator = limMaxInt;
+
+  } else if (integrator < limMinInt) {
+    integrator = limMinInt;
+  }
+
+  // Derivative Term (optional) with low pass filter
+  differentiator
+    = -(2.0f * Kd * (measurement - prevMeasurement) /* Note: derivative on measurement, therefore
+                                                       minus sign in front of equation! */
```
TomLonergan03, line 36: this comment is kinda ugly in the way it forces a linebreak into the equation, should prob just be a one line above the equation

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+        + (2.0f * tau - T) * differentiator)
+      / (2.0f * tau + T);
+
+  // Set controller output and limit outputs if needed
+  out = proportional + integrator + differentiator;
```
TomLonergan03, line 41: to clarify what i said in the header, declare this here:
```c++
const core::Float output = ...
```

lib/motors/PID.cpp
```diff
@@ -0,0 +1,56 @@
...
+  if (out > limMax) {
+    out = limMax;
+
+  } else if (out < limMin) {
+    out = limMin;
+  }
+
+  // Store error and measurement for later use
+  prevError       = error;
+  prevMeasurement = measurement;
```
TomLonergan03, line 52: do this before calculating the output, then you are able to refactor to get:
```c++
previous_error_ = error;
previous_measurement_ = measurement;
const core::Float output = ...;
if (output < minimum_output_) {
  return minimum_output_;
}
if (output > maximum_output_) {
  return maximum_output_;
}
return output;
```

lib/motors/PID.cpp
TomLonergan03, line 1: both files should match the name of the class they implement, so pid_controller.[ch]pp

lib/motors/PID.cpp
Hiatus44I: The values we want to configure:

`kp_, ki_, kd_, tau_, minimum_output_, maximum_output_, minimum_integrator_, maximum_integrator_, sample_time_` 

are all `const core::double` in the hpp file. Should I still initialise them in the constructor in the cpp file? Also where do i initialise `integrator_` and `previous_error_` etc? 

And to check, where should `kp` be `kp_` in the code (same with others)? Is it only the data members in the hpp class that should have the trailing dash?

lib/motors/PID.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+class PIDController {
+ public:
+  void PIDController_Init();
+  float PIDController_Update(float setpoint, float measurement);
```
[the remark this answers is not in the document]
Hiatus44I: When adding `#include <core/types.hpp>` I get an error saying it cannot open the source file. I think it's something to do with the path.

lib/motors/PID.hpp
```diff
@@ -0,0 +1,36 @@
...
+  // Output limits
+  double limMin;
+  double limMax;
+
+  // Integrator limits
+  double limMinInt;
+  double limMaxInt;
+
+  /* Sample time (in seconds) */
+  double T;
```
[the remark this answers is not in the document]
Hiatus44I: To clarify, it should be `const core::double ...`?

lib/motors/PID.cpp
Hiatus44I: This was also initialised to zero in what was `PIDController_Init` and is now the constructor `PidController`. Do I declare and initialise there? This kind of links to my previous question above.

lib/motors/PID.hpp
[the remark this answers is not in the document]
TomLonergan03: is that just in VSCode or also when you build it? vscode is sometimes weird about includes, there are solutions online but i've never completely got them to go away

lib/motors/PID.hpp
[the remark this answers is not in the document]
TomLonergan03: yep, `const core::Float`. `const` marks them as not changing after the class is constructed

lib/motors/PID.cpp
TomLonergan03: 1. yes, you can do that in the initaliser list in the cpp file like `kp_(kp)`
2. `integrator_` should also be initialised in the list, but you can just initialise it to 0 instead of a constructor argument
3. `kp` is used only in the constructor to initialise `kp_`, after that throughout the class you use the class' data member so it is `kp_`

It might also be worth putting all of those into a `PidParameters` struct, just to make the constructor a bit cleaner

lib/motors/PID.cpp
TomLonergan03: yeah, instead of making it a class data member you can just declare it on each iteration here, cos i don't believe you need the output value to persist between calls of `update(...)`

lib/motors/PID.cpp
TomLonergan03, line 1: delete these files now you've renamed the other ones

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+      previous_measurement_{0},
+      kp_(kp),
+      ki_(ki),
+      kd_(kd),
+      tau_(tau),
+      minimum_output_(minimum_output),
+      maximum_output_(maximum_output),
+      minimum_integrator_(minimum_integrator),
+      maximum_integrator_(maximum_integrator),
+      sample_time_(sample_time){};
```
TomLonergan03, line 27: no `;` needed after implementations of methods defined in the header

lib/motors/pid_controller.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  // Output limits
+  const core::Float minimum_output_;
+  const core::Float maximum_output_;
+
+  // Integrator limits
+  const core::Float minimum_integrator_;
+  const core::Float maximum_integrator_;
+
+  /* Sample time (in seconds) */
```
TomLonergan03, line 38: one line comments with `//`

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+core::Float PidController::update(core::Float setpoint, core::Float measurement)
+{
+  const core::Float error = setpoint - measurement;
+
```
TomLonergan03, line 32: still a whole bunch of whitespace, i think it all can go

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+  const core::Float proportional = kp_ * error;
+
+  integrator_ = integrator_ + 0.5f * ki_ * sample_time_ * (error + previous_error_);
```
TomLonergan03, line 35: you don't need to say `0.5f`, `0.5` is fine as everything is a double here

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+  if (integrator_ > maximum_integrator_) {
+    integrator_ = maximum_integrator_;
+
+  } else if (integrator_ < minimum_integrator_) {
+    integrator_ = minimum_integrator_;
+  }
+
+  // Derivative term (optional) with low pass filter
+  differentiator_
+    // Derivative on measurement, hence minus sign in front of equation
```
TomLonergan03, line 47: up another line, with the other comment

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+  if (output > maximum_output_) {
+    output = maximum_output_;
+
+  } else if (output < minimum_output_) {
+    output = minimum_output_;
+  }
+
+  // Return controller output
+  return output;
+};
```
TomLonergan03, line 68: `;` not needed

lib/motors/pid_controller.hpp
```diff
@@ -0,0 +1,47 @@
+#include "core/types.hpp"
+#pragma once
+
+namespace hyped::motors {
+class PidController {
+ public:
+  PidController(
+
```
TomLonergan03, line 8: whitespace, i'm suprised clang-format is alright with this i thought it would change it ngl

TomLonergan03 (changes requested):
we're getting close

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+core::Float PidController::update(core::Float setpoint, core::Float measurement)
+{
+  const core::Float error        = setpoint - measurement;
+  const core::Float proportional = kp_ * error;
+  integrator_ = integrator_ + 0.5 * ki_ * sample_time_ * (error + previous_error_);
+
+  // Integrator anti wind-up (dynamic integrator clamping)
+  if (integrator_ > maximum_integrator_) {
+    integrator_ = maximum_integrator_;
+
```
TomLonergan03, line 40: whitespace

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+  } else if (integrator_ < minimum_integrator_) {
+    integrator_ = minimum_integrator_;
+  }
+
+  // Derivative term (optional and on measurement, hence minus sign) with low pass filter
+  differentiator_ = -(2.0f * kd_ * (measurement - previous_measurement_)
+                      + (2.0f * tau_ - sample_time_) * differentiator_)
+                    / (2.0f * tau_ + sample_time_);
+
+  // Store error and measurement for later use
```
TomLonergan03, line 50: unnecessary comment

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+  previous_error_       = error;
+  previous_measurement_ = measurement;
+
+  // Set controller output and limit outputs if needed
```
TomLonergan03, line 54: remove comment + whitespace

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+  core::Float output = proportional + integrator_ + differentiator_;
+
+  if (output > maximum_output_) {
+    output = maximum_output_;
+
```
TomLonergan03, line 59: whitespace

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+  } else if (output < minimum_output_) {
+    output = minimum_output_;
+  }
+
+  // Return controller output
```
TomLonergan03, line 64: unnecessary comment

lib/motors/pid_controller.hpp
```diff
@@ -0,0 +1,44 @@
+#include "core/types.hpp"
+#pragma once
```
TomLonergan03, line 2: `pragma` above the includes, add whitespace between `pragma`, `includes`, and `namespace`

lib/motors/pid_controller.hpp
```diff
@@ -0,0 +1,44 @@
...
+  const core::Float maximum_integrator_;
+
+  // Sample time (in seconds)
+  const core::Float sample_time_;
+
+  /* Controller memory */
+  core::Float integrator_;
+  core::Float previous_error_; /* Required for integrator */
+  core::Float differentiator_;
+  core::Float previous_measurement_; /* Required for differentiator */
```
TomLonergan03, line 42: One line comments should be `//`

TomLonergan03 (changes requested):
Fix these and you should be good to merge it

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,67 @@
+#include "pid_controller.hpp"
+
+namespace hyped::motors {
+PidController::PidController(
+
```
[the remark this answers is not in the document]
Hiatus44I: For all these whitespaces, I've changed it but clang-format on VS doesn't seem to like it

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,55 @@
...
+  // Integrator anti wind-up (dynamic integrator clamping)
+  if (integrator_ > maximum_integrator_) {
+    integrator_ = maximum_integrator_;
+  } else if (integrator_ < minimum_integrator_) {
+    integrator_ = minimum_integrator_;
+  }
+  // Derivative term (optional and on measurement, hence minus sign) with low pass filter
+  differentiator_ = -(2.0f * kd_ * (measurement - previous_measurement_)
+                      + (2.0f * tau_ - sample_time_) * differentiator_)
+                    / (2.0f * tau_ + sample_time_);
```
TomLonergan03, line 43: don't need the `f`s

lib/motors/pid_controller.cpp
```diff
@@ -0,0 +1,55 @@
...
+                             const core::Float tau,
+                             const core::Float minimum_output,
+                             const core::Float maximum_output,
+                             const core::Float minimum_integrator,
+                             const core::Float maximum_integrator,
+                             const core::Float sample_time)
+    : integrator_{0},
+      previous_error_{0},
+      differentiator_{0},
+      previous_measurement_{0},
```
TomLonergan03, line 16: round brackets, `(0)`

TomLonergan03 (changes requested):
Last things

TomLonergan03 (approved):
Looks good! Thank you for making all the changes so quickly

DylanCavers (approved):
lgtm

test/motors/pid_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+
+
+
+TEST(motors, pid)
+{
+
+    PidController pidcontroller(2, 2, 2, 2, -19802, 5, -19404, 5, 1);
+
+    float measurement_list[100] ={
```
TomLonergan03, line 13: this should be a C++ array `std::array<core::Float>`

test/motors/pid_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+        2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,
+        42,44,46,48,50,52,54,56,58,60,62,64,66,68,70,72,74,76,78,80,
+        82,84,86,88,90,92,94,96,98,100,102,104,106,108,110,112,114,116,
+        118,120,122,124,126,128,130,132,134,136,138,140,142,144,146,148,
+        150,152,154,156,158,160,162,164,166,168,170,172,174,176,178,180,
+        182,184,186,188,190,192,194,196,198,200};
+
+    float expected_output[100] ={
```
TomLonergan03, line 21: same for this

test/motors/pid_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+-17858.00000000000,
+-18238.00000000000,
+-18622.00000000000,
+-19010.00000000000,
+-19402.00000000000,
+-19798.00000000000
+
+
+};
+    for (int i = 0; i < 100; i++){
```
TomLonergan03, line 125: `std::size_t` as the type of `i`

TomLonergan03 (changes requested):
`test/motors/pid_test.cpp` will also needs a `CMakeLists.txt` which will be something like
```cmake
set(target hyped_test_motors)
file(GLOB code "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
add_executable(${target} ${code})
target_include_directories(${target}
    PRIVATE
    "${CMAKE_SOURCE_DIR}/lib"
    "${CMAKE_CURRENT_SOURCE_DIR}"
)
target_link_libraries(${target}
    GTest::gtest_main
    hyped_core
    hyped_motors
)
gtest_discover_tests(${target})
```

TomLonergan03:
Also run `clang-format` over this

TomLonergan03 (approved):

PR #20, opened 2023-10-28
MGT - Fix github CI

TomLonergan03:
Due to https://github.com/actions/runner-images/issues/8659, we have to manually install and use clang-17 or every `#include <chrono>` breaks. This will be removed once either the issue is fixed or I set up the docker build image.

It also adds ~30s to the linux build, sorry about that


kshxtij (approved):
LGTM, merge when the CI completes 👍

DylanCavers (approved):

PR #21, opened 2023-10-31
TEL - Docker!

davidbeechey:
In a staggering feat of procrastination, I have finally dockerized telemetry 🥳 (I think)

Just run `docker-compose up --build` in `/telemetry` and it'll spin up all of the telemetry services, mosquitto (MQTT broker) and InfluxDB (the time-series database we use for saving historical readings) all fully configured. The GUI should then be available at `localhost:5173`, MQTT broker on `localhost:1883` and the InfluxDB management dashboard at `localhost:8086`.

### DIFF
--- a/telemetry/Dockerfile
+++ b/telemetry/Dockerfile
@@ -1,0 +1,31 @@
+FROM node:18.17.0
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install dependencies
+COPY package.json ./
+COPY packages/constants/package.json ./packages/constants/
+COPY packages/server/package.json ./packages/server/
+COPY packages/types/package.json ./packages/types/
+COPY packages/ui/package.json ./packages/ui/
+COPY yarn.lock ./
+COPY patches ./patches
+RUN yarn install
+
+# Copy all source files
+COPY . .
+
+# Replace server/ui .env with .env.docker
+RUN cp packages/server/.env.docker packages/server/.env
+RUN cp packages/ui/.env.docker packages/ui/.env
+
+# Build app
+RUN yarn build
+
+# Expose ports
+EXPOSE 5173
+EXPOSE 3000
+
+# Start app
+CMD [ "yarn", "dev" ]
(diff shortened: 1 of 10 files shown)


telemetry/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM node:18.17.0
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install dependencies
+COPY package.json ./
```
kshxtij, line 7: Depending on your workflow, I would prefer mounting the entire telemetry folder as a volume into `/usr/src/app` instead of copying files over on container build which would require rebuilding for every change.

kshxtij (changes requested):
Just a small suggestion/question but otherwise LGTM.

telemetry/Dockerfile
davidbeechey: Good idea, will do. Thanks :)

TomLonergan03 (commented):
image
is this me or you?

all i have at `packages/server` is `.env.example`

davidbeechey:
> image is this me or you?
> 
> all i have at `packages/server` is `.env.example`

I've included the `.env.docker` file now so should work?

TomLonergan03 (approved):
Do Kshitij's suggestion, otherwise looks good.

Is there anything to change with the github CI now we have this?

DylanCavers (approved):

TomLonergan03 (changes requested):
Update the readme with the new launch process also

kshxtij:
Im not sure if this works as everything just exits 🤔 

Screenshot 2023-11-22 at 22 04 40

davidbeechey:
**Update:** New way to run is: `./docker.sh <yarn_script>`.

TomLonergan03 (approved):
LGTM

kshxtij:
Wasn't able to test locally, but @TomLonergan03 feel free to merge if you were able to run it locally

kshxtij (approved):
Works locally 👍, Good job @davidbeechey

PR #25, opened 2023-11-06
Add files via upload

RheaBose:


lib/sensors/tof
```diff
@@ -0,0 +1,40 @@
...
+  }
+  logger.log(
+    core::LogLevel::kDebug, "Successful to configure VL6180X sensor at channel %d", channel);
+  return VL6180X(logger, i2c, channel, device_address);
+}
+
+
+
+
+}
```
DylanCavers, line 40: namespace comment needed here. I also think this file isn't a `.cpp` file, so that should probably be changed

lib/sensors/tof
```diff
@@ -0,0 +1,40 @@
+#include <iostream>
+#include<.clang-format>
+#include<i2c_sensors.hpp>
+#include "tof.hpp"
+
+namespace hyped::sensor{
+std::optional<TimeOfFlight> TimeOfFlight::create(core::ILogger &logger,
+                                                       std::shared_ptr<io::II2c> i2c,
+                                                       const std::uint8_t device_address)
+{ {
```
DylanCavers, line 10: two opening brackets?

lib/sensors/tof
```diff
@@ -0,0 +1,40 @@
...
+  return TimeOfFlight(logger, i2c, device_address);
+}
+
+TimeOfFlight::TimeOfFlight    (core::ILogger &logger,
```
DylanCavers, line 14: some of this could be formatted a bit better, make sure you are using `clang-format` to format all your `.cpp` and `.hpp` files

lib/sensors/tof
```diff
@@ -0,0 +1,40 @@
...
+    return std::nullopt;
+  }
+  logger.log(
+    core::LogLevel::kDebug, "Successful to configure VL6180X sensor at channel %d", channel);
+  return VL6180X(logger, i2c, channel, device_address);
+}
+
+
+
+
```
DylanCavers, line 39: remove the spaces on these lines just to make it look nicer

lib/sensors/tof.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include "i2c_sensors.hpp"
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>  
+
+namespace hyoed :: sensors {
```
DylanCavers, line 12: this needs to be `hyped::sensors`

lib/sensors/tof.hpp
```diff
@@ -0,0 +1,49 @@
...
+          const std::uint8_t channel,
+          const std::uint8_t device_address);
+
+ private:
+  core::ILogger &logger_;
+  std::shared_ptr<io::II2c> i2c_;
+  const std::uint8_t channel_;
+  const std::uint8_t device_address_;
+
+private:
```
DylanCavers, line 39: there should only be one private in the class definition, so should remove the duplicated ones

lib/sensors/tof
```diff
@@ -0,0 +1,40 @@
...
+}
+
+TimeOfFlight::TimeOfFlight    (core::ILogger &logger,
+                                 std::shared_ptr<io::II2c> i2c,
+                                 const std::uint8_t device_address)
+    : logger_(logger),
+      i2c_(i2c),
+      device_address_(device_address)
+{
+{
```
DylanCavers, line 21: remove one of these brackets too

DylanCavers (commented):
A few changes, mostly to do with formatting and code style, make sure to change the tof file to `tof.cpp`

lib/sensors/tof
TomLonergan03, line 1: this should be `time_of_flight.cpp`

lib/sensors/tof.hpp
TomLonergan03, line 1: `time_of_flight.hpp`

lib/sensors/tof
```diff
@@ -0,0 +1,40 @@
+#include <iostream>
+#include<.clang-format>
```
TomLonergan03, line 2: i don't think this should be included

TomLonergan03 (changes requested):
also, run clang-format over this

TomLonergan03 (changes requested):
Rename this PR also

sa-fx:
Implemented with reference to STMicroelectronics application note AN4545

- [ ] Rebase with `master` after #58  has been merged
- [ ] Confirm requirements from Electronics for what this sensor should do (Is ambient light reading required? Continuous or single-shot?)
- [ ] Implement initialisation of ToF, including registers without names
- [x] Implement reading continuous range
- [x] Implement reading single-shot range
- [ ] Implement `II2cMuxSensor`
- [ ] Test on hardware

lib/sensors/time_of_flight.hpp
```diff
@@ -0,0 +1,102 @@
...
+   * @brief Check if the sensor is in the appropriate mode before trying to read the range, and set
+   * it to mode_value if it isn't
+   * @param mode_value  0b01 or 0b11 for single-shot or continuous
+   * @return kSuccess if the sensor mode is correct, and kFailure otherwise
+   */
+  core::Result checkSensorMode(std::uint8_t mode_value);
+
+ private:
+  // TODOLater - Confirm these addresses are correct
+  // Register addresses/values taken from the VL6180X datasheet
```
sa-fx, line 65: Not sure where exactly they've been taken from, or what they all mean - Are these needed?

lib/sensors/time_of_flight.cpp
```diff
@@ -0,0 +1,180 @@
...
+    // i2c_->writeByteToRegister(device_address_, 0x0198, 0x01);
+    // i2c_->writeByteToRegister(device_address_, 0x01b0, 0x17);
+    // i2c_->writeByteToRegister(device_address_, 0x01ad, 0x00);
+    i2c_->writeByteToRegister(device_address_, 0x00ff, 0x05);
+    // i2c_->writeByteToRegister(device_address_, 0x0100, 0x05);
+    // i2c_->writeByteToRegister(device_address_, 0x0199, 0x05);
+    // i2c_->writeByteToRegister(device_address_, 0x01a6, 0x1b);
+    // i2c_->writeByteToRegister(device_address_, 0x01ac, 0x3e);
+    // i2c_->writeByteToRegister(device_address_, 0x01a7, 0x1f);
+    i2c_->writeByteToRegister(device_address_, 0x0030, 0x00);
```
TomLonergan03, line 77: this whole thing is gross but there's probably no better way

lib/sensors/time_of_flight.cpp
```diff
@@ -0,0 +1,180 @@
...
+{
+  if (checkSensorMode(kModeSingleShot) == core::Result::kFailure) { return std::nullopt; }
+  return getRange();
+};
+
+std::optional<std::uint8_t> TimeOfFlight::getRangeContinuous()
+{
+  if (checkSensorMode(kModeContinuous) == core::Result::kFailure) { return std::nullopt; }
+  return getRange();
+};
```
TomLonergan03, line 107: what are these for?

lib/sensors/time_of_flight.cpp
```diff
@@ -0,0 +1,180 @@
...
+      return core::Result::kFailure;
+    }
+  }
+  return core::Result::kSuccess;
+};
+
+std::optional<std::uint8_t> TimeOfFlight::getRange()
+{
+  // Check the status
+  auto status = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
```
TomLonergan03, line 130: `const`

lib/sensors/time_of_flight.hpp
```diff
@@ -0,0 +1,102 @@
...
+  static constexpr std::uint8_t kModeSingleShot = 0b01;
+  static constexpr std::uint8_t kModeContinuous = 0b11;
+
+  // TODOLater - std::uint8_t or std::uint16_t for all regs?
+  static constexpr std::uint8_t kSystemFreshOutOfReset  = 0x016;
+  static constexpr std::uint8_t kSystemModeGpioOne      = 0x011;
+  static constexpr std::uint16_t kReadoutSamplingPeriod = 0x010a;
+  static constexpr std::uint8_t kSysAlsAnalogueGain     = 0x03f;
+  static constexpr std::uint8_t kSysRangeVhvRepeatRate  = 0x031;
+  // TODOLater - Confirm SYSALS__INTEGRATION_PERIOD (application note vs datasheet)
```
sa-fx, line 82: The datasheet gives this register an address of `0x040`, but the application note AN4545 gives this register an address of `0x0041` on page 24

lib/sensors/time_of_flight.cpp
```diff
@@ -0,0 +1,180 @@
...
+  if (!status) {
+    logger_.log(core::LogLevel::kFatal, "Failed to get time of flight interrupt status");
+    return std::nullopt;
+  }
+  auto range_status = *status & 0b111;
+
+  // Wait for new measurement ready status
+  while (range_status != 0x04) {
```
TomLonergan03, line 138: this is scary, we need to be able to guarantee that it can't get stuck in this loop

TomLonergan03 (changes requested):
ik it's a wip, but here's critique anyway lol

lib/sensors/time_of_flight.hpp
TomLonergan03: lmfao of course it does

engineers write a useful datasheet challenge (impossible)

lib/sensors/time_of_flight.cpp
sa-fx: The sensor has two modes of operation: Single shot, and continuous. Single shot is a single measurement, while continuous will repeatedly measure the range with a time interval based on the value of `kSysRangeIntermeasurementPeriod`. Before reading in a given mode, register `kSysRangeStart` has to be set to either 1 or 3
Because the code to check that `kSysRangeStart` has the correct value is the same whether its checking for continuous or single-shot, and because the register to read the range from is the same in both cases if `kSysRangeStart` has the correct value (`kResultRangeVal`), `getRangeSingleShot()` and `getRangeContinuous()` call these separate functions to retrieve the range

As for which value will be wanted, single shot or continous, that's yet to be decided

lib/sensors/time_of_flight.cpp
```diff
@@ -0,0 +1,181 @@
...
+    i2c_->writeByteToRegister(device_address_, kSystemModeGpioOne, 0x10);
+    // i2c_->writeByteToRegister(device_address_, kReadoutSamplingPeriod, 0x30);
+    i2c_->writeByteToRegister(device_address_, kSysAlsAnalogueGain, 0x46);
+    i2c_->writeByteToRegister(device_address_, kSysRangeVhvRepeatRate, 0xff);
+    i2c_->writeByteToRegister(device_address_, kSysAlsIntegrationPeriod, 0x63);
+    i2c_->writeByteToRegister(device_address_, kSysRangeVhvRecalibrate, 0x01);
+
+    // Optional: Public registers
+    i2c_->writeByteToRegister(device_address_, kSysRangeIntermeasurementPeriod, 0x09);
+    i2c_->writeByteToRegister(device_address_, kSysAlsIntermeasurementPeriod, 0x31);
```
TomLonergan03, line 92: make an array of values for all of these and loop through it

lib/sensors/tof.cpp
```diff
@@ -48,40 +48,32 @@ core::Result TimeOfFlight::initialise()
...
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[9], 0xce);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[10], 0x03);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[11], 0xf8);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[12], 0x00);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[13], 0x3c);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[14], 0x00);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[15], 0x3c);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[16], 0x09);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[17], 0x09);
-    // i2c_->writeByteToRegister(device_address_, 0x0198, 0x01);
```
sa-fx: These registers were not commented out because they should be removed. They were commented out because at the time this code was written, #58 hadn't been merged yet. As such, the 16-bit registers being written to would cause build failure.
These registers should be added back in. Refer to the application note in the PR's header

sa-fx: These registers were not commented out because they should be removed. They were commented out because at the time this code was written, #58 hadn't been merged yet. As such, the 16-bit registers being written to would cause build failure.
These registers should be added back in. Refer to the application note in the PR's header

lib/sensors/tof.cpp
```diff
@@ -128,21 +120,28 @@ core::Result TimeOfFlight::checkSensorMode(std::uint8_t mode_value)
...
     return std::nullopt;
   }
   auto range_status = *status & 0b111;
 
   // Wait for new measurement ready status
+  uint8_t timeout = 0;
   while (range_status != 0x04) {
-    status = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
-    if (!status) {
+    const auto statuss = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
```
sa-fx, line 133: `statuss` is quite the variable name.

lib/sensors/tof.cpp
```diff
@@ -128,21 +120,28 @@ core::Result TimeOfFlight::checkSensorMode(std::uint8_t mode_value)
...
-  auto status = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
+  const auto status = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
   if (!status) {
     logger_.log(core::LogLevel::kFatal, "Failed to get time of flight interrupt status");
     return std::nullopt;
   }
   auto range_status = *status & 0b111;
 
   // Wait for new measurement ready status
+  uint8_t timeout = 0;
```
sa-fx, line 131: `std::uint8_t`

lib/sensors/tof.cpp
```diff
@@ -128,21 +120,28 @@ core::Result TimeOfFlight::checkSensorMode(std::uint8_t mode_value)
...
-    if (!status) {
+    const auto statuss = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
+    if (!statuss) {
       logger_.log(core::LogLevel::kFatal, "Failed to get time of flight interrupt status");
       return std::nullopt;
     }
-    range_status = *status & 0b111;
+    range_status = *statuss & 0b111;
+    timeout++;
+    if (timeout > 1000000) {
```
sa-fx, line 140: Comparing a `uint8_t` variable to 1,000,000 will always return false

sa-fx, line 140: Is 1,000,000 sensible? It seems quite large.

lib/sensors/tof.cpp
```diff
@@ -48,40 +48,32 @@ core::Result TimeOfFlight::initialise()
...
+                                     0x05,
+                                     0x00};
+    for (int i = 0; i < 20; i++) {
+      i2c_->writeByteToRegister(device_address_, array_return_addresses[1], data[i]);
+    }
     i2c_->writeByteToRegister(device_address_, kSystemFreshOutOfReset, 0);
 
     // Recommended : Public registers
     i2c_->writeByteToRegister(device_address_, kSystemModeGpioOne, 0x10);
-    // i2c_->writeByteToRegister(device_address_, kReadoutSamplingPeriod, 0x30);
```
sa-fx: Again, this was a 16-bit register which would cause build failure, not a line which needed to be removed. It should be added back in

lib/sensors/tof.cpp
```diff
@@ -48,40 +48,32 @@ core::Result TimeOfFlight::initialise()
...
-    // i2c_->writeByteToRegister(device_address_, 0x01ad, 0x00);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[18], 0x05);
-    // i2c_->writeByteToRegister(device_address_, 0x0100, 0x05);
-    // i2c_->writeByteToRegister(device_address_, 0x0199, 0x05);
-    // i2c_->writeByteToRegister(device_address_, 0x01a6, 0x1b);
-    // i2c_->writeByteToRegister(device_address_, 0x01ac, 0x3e);
-    // i2c_->writeByteToRegister(device_address_, 0x01a7, 0x1f);
-    i2c_->writeByteToRegister(device_address_, array_return_addresses[19], 0x00);
-
+    constexpr std::uint8_t data[] = {0xfd,
```
sa-fx, line 51: Add entries to this for the values to be written to the 16-bit registers

lib/sensors/tof.cpp
```diff
@@ -1,4 +1,4 @@
-#include "time_of_flight.hpp"
+#include "tof.hpp"
```
sa-fx, line 1: [Why was this changed back?](https://github.com/Hyp-ed/hyped-2024/pull/25#pullrequestreview-1723125739)

lib/sensors/tof.hpp
sa-fx, line 1: Make `array_return_addresses` a `std::uint16_t` array, and add the 16-bit registers. Then you can itterate through all the registers
(AN4545, p20)

sa-fx (changes requested):

lib/sensors/tof.cpp
sa-fx, line 1: You should delete these `tof.cpp/hpp` files.

lib/sensors/time_of_flight.cpp
```diff
@@ -0,0 +1,189 @@
...
+  const auto status = i2c_->readByte(device_address_, kResultInterruptStatusGpio);
+  if (!status) {
+    logger_.log(core::LogLevel::kFatal, "Failed to get time of flight interrupt status");
+    return std::nullopt;
+  }
+  auto range_status = *status & 0b111;
+
+  // Wait for new measurement ready status
+  std::uint8_t timeout = 0;
+  std::uint8_t threshold = 10000;
```
sa-fx, line 140: `std::unit8_t` has a max value of 255. If you want to use 10,000 here, you'll need `std::uint16_t`

sa-fx, line 140: Where has the value of 10,000 come from? What's the rationale for choosing that value?

PR #26, opened 2023-11-06
SNS - Minor changes in GPIO implementation

licornes-fluos:
Change for GPIO protocol, matched IGPIO and HardwareGpio, and implements IGPIO in HardwareGpio.

### DIFF
--- a/lib/io/gpio.hpp
+++ b/lib/io/gpio.hpp
@@ -39,9 +39,11 @@ class IGpioWriter {
 class IGpio {
  public:
   virtual std::optional<std::shared_ptr<IGpioReader>> getReader(const std::uint8_t pin,
-                                                                const Edge edge) = 0;
+                                                                const Edge edge)
+    = 0;
   virtual std::optional<std::shared_ptr<IGpioWriter>> getWriter(const std::uint8_t pin,
-                                                                const Edge edge) = 0;
+                                                                const Edge edge)
+    = 0;
 };
 
 }  // namespace hyped::io
--- a/lib/io/gpio.hpp
+++ b/lib/io/gpio.hpp
@@ -9,6 +9,10 @@
 
 namespace hyped::io {
 
+// Edge is used to set the interrupt trigger for the pin.
+enum class Edge { kNone = 0, kRising, kFalling, kBoth };
+enum class Direction { kIn = 0, kOut };
+
 /**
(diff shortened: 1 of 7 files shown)


lib/io/gpio.hpp
```diff
@@ -34,8 +34,10 @@ class IGpioWriter {
  */
 class IGpio {
  public:
-  virtual std::optional<std::shared_ptr<IGpioReader>> getReader(const std::uint8_t pin) = 0;
-  virtual std::optional<std::shared_ptr<IGpioWriter>> getWriter(const std::uint8_t pin) = 0;
+  virtual std::optional<std::shared_ptr<IGpioReader>> getReader(const std::uint8_t pin,
+                                                                const Edge edge) = 0;
+  virtual std::optional<std::shared_ptr<IGpioWriter>> getWriter(const std::uint8_t pin,
+                                                                const Edge edge) = 0;
```
TomLonergan03, line 40: this is the other thing, a bunch of stuff like `Edge` are defined in `hardware_gpio.hpp`, which i don't think should be there

lib/io/hardware_gpio.hpp
```diff
@@ -51,7 +51,7 @@ class HardwareGpioWriter : public IGpioWriter {
  * be instantiated at the top level and then provided to users through the IGpio interface.
  * Ensure inputted pins are defined as pin = 32*X + Y (GPIOX_Y)
  */
-class HardwareGpio {
+class HardwareGpio : public IGpio{
```
TomLonergan03, line 54:
```suggestion
class HardwareGpio : public IGpio {
```

TomLonergan03 (changes requested):

lib/io/gpio.hpp
licornes-fluos: Where would it be if not here though?

TomLonergan03: yes, those should be defined here as this doesn't build

TomLonergan03 (changes requested):
`clang-format` + fixes in keyence then good to go

TomLonergan03 (changes requested):
`DummyGpio` will need updated to match the interface changes also

davidbeechey (approved):

TomLonergan03 (approved):

PR #27, opened 2023-11-09
NAV - Transfer KF & EKF logic from header to .cpp

pkhudov:
EKF does not compile because dimensions of matrices do not match. Unsure why the error wasn't caught in .hpp before.


lib/navigation/filtering/kalman_filter.hpp
```diff
@@ -7,57 +7,44 @@
...
-}  // namespace hyped::navigation
+   inline const StateVector& KalmanFilter::getStateEstimate() const
+{
+  return state_estimate_;
+}
+
+   inline const ErrorCovarianceMatrix& KalmanFilter::getErrorCovariance() const
+{
+  return error_covariance_;
+}
```
TomLonergan03, line 45: These are declared twice

lib/navigation/filtering/extended_kalman_filter.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+ExtendedKalmanFilter::ExtendedKalmanFilter(std::shared_ptr<core::ITimeSource> time_source,
+                                           const StateVector initial_state,
+                                           const ErrorCovarianceMatrix initial_error_covariance)
+    : time_source_(time_source),
+      last_update_time_(time_source->now()),
+      state_estimate_(initial_state),
+      error_covariance_(initial_error_covariance)
+{
+  static_assert(state_dimension > 0);
```
kshxtij, line 13: HYPED standards have normally been to use a create function which returns an optional<class> for error handling in the constructor (most of the sensors are a good example)

lib/navigation/filtering/kalman_filter.cpp
```diff
@@ -0,0 +1,43 @@
...
+              const MeasurementVector &measurement)
+{
+      // TODOLater: figure out how to make transition matrix given time delta - in main nav section
+    const auto prop_state_estimate = transition_matrix * state_estimate_;
+    const auto prop_error_covariance
+      = (transition_matrix.transpose() * error_covariance_ * transition_matrix)
+        + transition_covariance;
+    // TODOLater: Some optimisation is to be found here:
+    // 1. Try and get rid of the inverse.
+    // 2. Reuse calculations such as `prop_error_covariance * measurement_matrix.transpose()`.
```
kshxtij, line 32: Can probably include this optimisation rn imo

kshxtij (changes requested):

lib/navigation/filtering/extended_kalman_filter.cpp
TomLonergan03: this is fine if we are defining the `state_dimension` in a `constexpr` then this catches that at compile time right? we would only need a `create` if these are in a config file, which idk enough about kalman filter dimensions to say if we want to do that

PR #29, opened 2023-11-09
STM - MQTT Functionality

DylanCavers:
Adding MQTT functionality to the state machine in order to receive transitions and publish the current state


test/state_machine/stm_tests.cpp
```diff
@@ -12,9 +15,26 @@ void testTransition(std::unique_ptr<state_machine::StateMachine> &stm,
   ASSERT_TRUE(stm->getCurrentState() == expected_state);
 }
 
+std::shared_ptr<core::Mqtt> getMqtt()
```
TomLonergan03, line 18: unless you have a way to use the broker as part of the tests, you're gonna want an `IMqtt` interface and a `MockMqtt` class in `lib/utils` I reckon

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +41,42 @@ core::Result StateMachine::handleTransition(const State &state)
...
+  rapidjson::Value state_value(state_string.c_str(), message_payload->GetAllocator());
+  message_payload->AddMember("transition", state_value, message_payload->GetAllocator());
+  core::MqttMessage::Header header{.timestamp = 0, .priority = core::MqttMessagePriority::kNormal};
+  const core::MqttMessage message{topic, header, message_payload};
+  mqtt_->publish(message, core::MqttMessageQos::kExactlyOnce);
+}
+
+void StateMachine::startStateMachine()
+{
+  while (true) {
```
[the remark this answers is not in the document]
DylanCavers: updated to finish when a shutdown command is received

lib/utils/dummy_mqtt.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+  return core::Result::kSuccess;
+}
+
+core::Result MockMqtt::consume()
+{
+  return core::Result::kSuccess;
+}
+
+std::optional<core::MqttMessage> MockMqtt::getMessage()
```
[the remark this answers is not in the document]
DylanCavers: will probably leave this till later when we think about testing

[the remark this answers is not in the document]
TomLonergan03: Fair enough

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +42,43 @@ core::Result StateMachine::handleTransition(const State &state)
...
+  mqtt_->publish(message, core::MqttMessageQos::kExactlyOnce);
+}
+
+void StateMachine::startStateMachine()
+{
+  while (StateMachine::getCurrentState() != State::kShutdown) {
+    auto updated = StateMachine::update();
+    StateMachine::publishCurrentState();
+  }
+  StateMachine::publishCurrentState();
```
TomLonergan03, line 81: these aren't static methods

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +42,43 @@ core::Result StateMachine::handleTransition(const State &state)
...
+  message_payload->AddMember("transition", state_value, message_payload->GetAllocator());
+  core::MqttMessage::Header header{.timestamp = 0, .priority = core::MqttMessagePriority::kNormal};
+  const core::MqttMessage message{topic, header, message_payload};
+  mqtt_->publish(message, core::MqttMessageQos::kExactlyOnce);
+}
+
+void StateMachine::startStateMachine()
+{
+  while (StateMachine::getCurrentState() != State::kShutdown) {
+    auto updated = StateMachine::update();
```
TomLonergan03, line 78: ??

if the return value matters we should do something with it, if not get rid of the return

lib/state_machine/transition_table.hpp
```diff
@@ -37,7 +37,8 @@ const TransitionTable transition_to_state_dynamic
      {{State::kBatteryRecharge, State::kFailure}, State::kFailure},
      {{State::kFailureBrake, State::kCapacitorDischarge}, State::kCapacitorDischarge},
      {{State::kFailure, State::kCapacitorDischarge}, State::kCapacitorDischarge},
-     {{State::kCapacitorDischarge, State::kSafe}, State::kSafe}};
+     {{State::kCapacitorDischarge, State::kSafe}, State::kSafe},
+     {{State::kSafe, State::kShutdown}, State::kShutdown}};
```
TomLonergan03, line 41: We probably want to shut down from some (I'm not sure which) pre-run states

TomLonergan03 (changes requested):

lib/state_machine/state_machine.cpp
DylanCavers: do we want them as static?

TomLonergan03: i mean you should construct a state machine and do `state_machine.update()`. I don't think these can be made static but they definitely should not be invoked as if they are

DylanCavers: nvm this was just a skill issue from me

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +42,38 @@ core::Result StateMachine::handleTransition(const State &state)
...
+  std::shared_ptr message_payload = std::make_shared<rapidjson::Document>();
+  message_payload->SetObject();
+  rapidjson::Value state_value(state_string.c_str(), message_payload->GetAllocator());
+  message_payload->AddMember("transition", state_value, message_payload->GetAllocator());
+  core::MqttMessage::Header header{.timestamp = 0, .priority = core::MqttMessagePriority::kNormal};
+  const core::MqttMessage message{topic, header, message_payload};
+  mqtt_->publish(message, core::MqttMessageQos::kExactlyOnce);
+}
+
+void StateMachine::startStateMachine()
```
TomLonergan03, line 70: `run()` imo

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +42,38 @@ core::Result StateMachine::handleTransition(const State &state)
   return core::Result::kFailure;
 }
 
+void StateMachine::update()
+{
+  mqtt_->consume();
```
TomLonergan03, line 47: is it cleaner to call this here or in `StateMachine::run()`? idk

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +42,38 @@ core::Result StateMachine::handleTransition(const State &state)
...
+
+void StateMachine::publishCurrentState()
+{
+  const auto state_string         = stateToString(getCurrentState());
+  const auto topic                = core::MqttTopic::kTest;
+  std::shared_ptr message_payload = std::make_shared<rapidjson::Document>();
+  message_payload->SetObject();
+  rapidjson::Value state_value(state_string.c_str(), message_payload->GetAllocator());
+  message_payload->AddMember("transition", state_value, message_payload->GetAllocator());
+  core::MqttMessage::Header header{.timestamp = 0, .priority = core::MqttMessagePriority::kNormal};
```
TomLonergan03, line 65: `const`

lib/state_machine/state_machine.hpp
```diff
@@ -8,22 +8,27 @@
 #include <queue>
 #include <unordered_map>
 
+#include "core/mqtt.hpp"
```
TomLonergan03, line 11: <angle brackets>

lib/state_machine/state_machine.hpp
```diff
@@ -58,6 +64,7 @@ class StateMachine {
        {State::kSafe, "kSafe"}};
 
   State current_state_;
+  std::shared_ptr<core::IMqtt> mqtt_;
   TransitionTable transition_to_state_;
```
TomLonergan03, line 68: `const`

TomLonergan03 (changes requested):

lib/state_machine/state_machine.cpp
```diff
@@ -35,4 +42,39 @@ core::Result StateMachine::handleTransition(const State &state)
...
+  const core::MqttMessage::Header header{.timestamp = 0,
+                                         .priority  = core::MqttMessagePriority::kNormal};
+  const core::MqttMessage message{topic, header, message_payload};
+  mqtt_->publish(message, core::MqttMessageQos::kExactlyOnce);
+}
+
+void StateMachine::run()
+{
+  while (getCurrentState() != State::kShutdown) {
+    mqtt_->consume();
```
TomLonergan03, line 73: this returns `kFailure` if it receives an invalid message, we should prob enter shutdown state in that case imo

TomLonergan03 (changes requested):
not much left

lib/state_machine/state_machine.cpp
DylanCavers: atm moment we can only transition to `kshutdown` from `kSafe`, should I add a transition from other states?

TomLonergan03 (approved):

lsalhani (approved):

PR #30, opened 2023-11-09
NAV - Fix accelerometer displacement estimate

pkhudov:
Added initial displacement to the position formula, now returning absolute track position instead of Δs. Otherwise, it would have never braked. Also got rid of velocity temp variable.

### DIFF
--- a/lib/navigation/preprocessing/accelerometer_trajectory.cpp
+++ b/lib/navigation/preprocessing/accelerometer_trajectory.cpp
@@ -23,24 +23,13 @@ void AccelerometerTrajectoryEstimator::update(const core::Float acceleration,
   const auto time_elapsed                = timer_.elapsed(timestamp, previous_timestamp_);
   const core::Float time_elapsed_seconds = timer_.elapsedTimeInSeconds(time_elapsed);
 
-  // from equation v=u+at
-  const core::Float velocity_estimate = velocity_estimate_ + (acceleration * time_elapsed_seconds);
-
-  // from equation s = ut + 0.5*a*(t^2)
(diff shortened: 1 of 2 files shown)


lib/navigation/preprocessing/accelerometer_trajectory.cpp
```diff
@@ -23,24 +23,13 @@ void AccelerometerTrajectoryEstimator::update(const core::Float acceleration,
...
-core::Float AccelerometerTrajectoryEstimator::getDisplacementEstimate() const
-{
-  return displacement_estimate_;
-}
 
-core::Float AccelerometerTrajectoryEstimator::getVelocityEstimate() const
-{
-  return velocity_estimate_;
+  // from equation v=u+at
+  velocity_estimate_  = velocity_estimate_ + (acceleration * time_elapsed_seconds);
```
TomLonergan03, line 31: should this be updated before or after the displacement estimate, as displacement uses this value

TomLonergan03 (changes requested):

lib/navigation/preprocessing/accelerometer_trajectory.cpp
pkhudov: It would make sense before, since the u in displacement is the initial velocity. I changed it in order not to use temp variable, if you look at previous code you'll see that it does same thing.

TomLonergan03: Oh my bad, I should have noticed that. All good then

TomLonergan03 (approved):

kshxtij (approved):
lmao, good catch!

PR #31, opened 2023-11-13
SNS - Software testing

HTTYDKing:
Testing whether the scripts written for the sensors would operate correctly.


lib/io/mock_hardware_i2c.cpp
TomLonergan03, line 1: these should be in utils tbh

TomLonergan03 (changes requested):
also, run clang-format over this

PR #44, opened 2023-11-23
SNS - LED Driver

H-Allen:


lib/sensors/led_driver.cpp
```diff
@@ -0,0 +1,97 @@
+#include "led_driver.hpp"
+
+namespace hyped::sensors {
+
+std::optional<LedDriver> LedDriver::create(core::ILogger &logger,
+                                           std::shared_ptr<io::II2c> i2c,
+                                           const std::uint8_t device_address)
+{
+  if (device_address != kDefaultLedDriverAddress) {
```
TomLonergan03, line 9: what's the point of taking an argument for the LED driver if it must match the default?

lib/sensors/led_driver.cpp
```diff
@@ -0,0 +1,97 @@
...
+    logger.log(core::LogLevel::kFatal, "Invalid device address for LED driver");
+    return std::nullopt;
+  }
+
```
TomLonergan03, line 13: avoid using whitespace (see the style guide [here](https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)). the metaphor is that whitespace in code is similar to whitespace in an essay, it should separate paragraphs but you wouldn't have a line break after every sentence

lib/sensors/led_driver.cpp
```diff
@@ -0,0 +1,97 @@
...
+      i2c_(i2c),
+      device_address_(device_address)
+{
+}
+
+LedDriver::~LedDriver()
+{
+}
+
+std::optional<core::Result> LedDriver::initialise()
```
TomLonergan03, line 38: this doesn't need to be `optional`, just return `core::Result::kFailure`/`core::Result::kSuccess`

TomLonergan03: same for all the others that currently return `std::optional<core::Result>`

lib/sensors/led_driver.cpp
```diff
@@ -0,0 +1,97 @@
...
+    return std::nullopt;
+  }
+
+  return core::Result::kSuccess;
+}
+
+std::optional<core::Result> LedDriver::set_colour(std::uint8_t channel,
+                                                  std::uint8_t red,
+                                                  std::uint8_t green,
+                                                  std::uint8_t blue)
```
TomLonergan03, line 54: all these args should be `const`

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "i2c_sensors.hpp"
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+int kDefaultLedDriverAddress = 0x30;
```
TomLonergan03, line 12: this should be `static constexpr std::uint8_t`

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace hyped::sensors {
+class LedDriver {
+ public:
+  static std::optional<LedDriver> create(core::ILogger &logger,
+                                         std::shared_ptr<io::II2c> i2c,
+                                         const std::uint8_t device_address
+                                         = kDefaultLedDriverAddress);
```
TomLonergan03, line 20: we don't use default parameters, but here we don't need the param at all so it's a kinda meaningless point to make

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,47 @@
...
+  ~LedDriver();
+
+  std::optional<core::Result> initialise();
```
TomLonergan03, line 23: this should be private i think

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,47 @@
...
+  std::optional<core::Result> set_colour(std::uint8_t channel,
```
TomLonergan03, line 24: functions names should be in `lowerCamelCase`

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,47 @@
...
+                                         std::uint8_t red,
+                                         std::uint8_t green,
+                                         std::uint8_t blue);
+  std::optional<core::Result> set_intensity(std::uint8_t channel, std::uint8_t intensity);
+  std::optional<core::Result> reset();
```
TomLonergan03, line 29: can you write a docstring for all of these methods, especially reset as it is a little unclear what it does

TomLonergan03 (changes requested):
looking good, mostly stylistic changes

lib/sensors/led_driver.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  logger.log(core::LogLevel::kDebug, "Successfully initialised LED driver");
+
+  return ledDriver;
+}
+
+LedDriver::LedDriver(core::ILogger &logger, std::shared_ptr<io::II2c> i2c)
+    : logger_(logger),
+      i2c_(i2c)
```
DylanCavers, line 20: might need a `std::move` here?

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+  core::Result initialise();
+  core::Result setColour(std::uint8_t channel,
+                         std::uint8_t red,
+                         std::uint8_t green,
+                         std::uint8_t blue);
+  core::Result setIntensity(std::uint8_t channel, std::uint8_t intensity);
+  core::Result reset();
+
+ private:
```
DylanCavers, line 28: don't need multiple `private`'s

lib/sensors/led_driver.hpp
```diff
@@ -0,0 +1,42 @@
...
+  LedDriver(core::ILogger &logger, std::shared_ptr<io::II2c> i2c);
```
DylanCavers, line 29: Constructor should be public

DylanCavers: although I'm not sure it needs to be if it is only called from create

lib/sensors/led_driver.hpp
TomLonergan03: at some point someone decided we have a second one separating variables from methods, idk if i like it tho

lib/sensors/led_driver.hpp
TomLonergan03: i feel like it has had to be public even with the create, but i dont remembeer why. if it builds it builds

lib/sensors/led_driver.cpp
```diff
@@ -0,0 +1,87 @@
...
+namespace hyped::sensors {
+
+std::optional<LedDriver> LedDriver::create(core::ILogger &logger, std::shared_ptr<io::II2c> i2c)
+{
+  auto ledDriver = LedDriver(logger, i2c);
+  if (auto init_result = ledDriver.initialise(); init_result == core::Result::kFailure) {
+    logger.log(core::LogLevel::kFatal, "Failed to initialise LED driver");
+    return std::nullopt;
+  }
+
```
TomLonergan03, line 12: remove the whitespace, same throughout the PR

TomLonergan03 (changes requested):

PR #46, opened 2023-11-25
STM - Static stm implementation

DH254:
Implementation do allow both a static and dynamic pod run using transitions as parameter

### DIFF
--- a/lib/state_machine/state_machine.cpp
+++ b/lib/state_machine/state_machine.cpp
@@ -4,7 +4,9 @@
 
 namespace hyped::state_machine {
 
-StateMachine::StateMachine() : current_state_{State::kIdle}
+StateMachine::StateMachine(const TransitionTable &transition_table) 
+  : current_state_{State::kIdle},
+  transition_to_state_{transition_table}
 {
 }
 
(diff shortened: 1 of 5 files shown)


TomLonergan03 (approved):

lib/state_machine/state_machine.cpp
```diff
@@ -4,7 +4,9 @@
 
 namespace hyped::state_machine {
 
-StateMachine::StateMachine() : current_state_{State::kIdle}
+StateMachine::StateMachine(const TransitionTable &transition_table)
+    : current_state_{State::kIdle},
```
DylanCavers, line 8: This was my mistake a while ago, quick fix here
```suggestion
    : current_state_(State::kIdle),
```

lib/state_machine/state_machine.cpp
```diff
@@ -4,7 +4,9 @@
...
+      transition_to_state_{transition_table}
```
DylanCavers, line 9:
```suggestion
      transition_to_state_(transition_table)
```

DylanCavers (commented):
will just make these changes then looks good

DylanCavers (approved):
lovely stuff

PR #47, opened 2023-11-25
Infra - Docker build

davidbeechey:


TomLonergan03 (commented):
What I think we do here is update the container so that we can also run the code within it, which fixes the linking issues for everyone. This then gets merged, and we start straight away on the crosscompilation extension

DylanCavers:
a few things to do before this can be merged:
- the dev environment does not build the image from scratch if it is not built, this is only done in the `build.sh` file. Should I add building the image to the dev file too? It would probably be the best idea as we wouldn't really use the build script for things other than the build.
- Make sure it actually supports mqtt, as we haven't tested this with the changes made on `infra-mqtt_tests` yet.

TomLonergan03:
> the dev environment does not build the image from scratch if it is not built, this is only done in the `build.sh` file. Should I add building the image to the dev file too? It would probably be the best idea as we wouldn't really use the build script for things other than the build.

Yeah, both should do that I think

TomLonergan03:
This works, any idea why the CLI doesn't do newlines between lines? Maybe it's just an issue on my end
image

davidbeechey:
> > This works, any idea why the CLI doesn't do newlines between lines? Maybe it's just an issue on my end image
> 
> I wasn't seeing this, so might just be on your end?

Same here... Arch problems? 😉

docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:20.04
```
DylanCavers, line 1: Also now that we aren't worried about syncing ubuntu version with wsl do we want this changed back to `latest`?

docker.sh
```diff
@@ -0,0 +1,113 @@
...
+if [ "$docker_dev" = true ]; then
+  # Check if the container name already exists
+  dev_container=$( docker ps -a -q --filter name=$DEV_CONTAINER_NAME 2> /dev/null )
+
+  if [[ -n ${dev_container} ]]; then
+    echo "[>] Found existing container"
+    docker start -i $DEV_CONTAINER_NAME
+  else
+    echo "[!] No existing container found. Creating new container"
+    docker run -it --mount type=bind,source=$(pwd),target=/home/hyped/ --name $DEV_CONTAINER_NAME --entrypoint /bin/bash $IMAGE_NAME
```
DylanCavers, line 111: should I add the cross-compile and clean stuff to the dev environment? and are we happy with the approach of starting an existing container if it exists rather than removing old containers and creating a new container each time (as is done for the build)?

docker/Dockerfile
TomLonergan03: might as well

docker/entry.sh
```diff
@@ -0,0 +1,19 @@
...
+if [[ -d "build" && $CLEAN = true ]]; then
+    rm -r build  
+    mkdir build
+elif [ ! -d "build" ]; then
+    mkdir build
+fi
+
+cd build 
+cmake ..
+make -j
```
TomLonergan03, line 19: can you also run `make test` here

docker.sh
TomLonergan03: I don't think we need cross compile in the dev env? I'm not sure. What is the practical difference between creating a new container or reusing an old one?

TomLonergan03:
> > This works, any idea why the CLI doesn't do newlines between lines? Maybe it's just an issue on my end image
> 
> I wasn't seeing this, so might just be on your end?

This fixed itself so IDK what was going on

docker.sh
DylanCavers: i think for the env it is fine to reuse, idk what difference it makes for the build

TomLonergan03: is it just that it doesn't preserve internal state between running? in that case we want a new container each time for builds, but the dev one idk. If it's mounting all the files from disk then there isn't really state and it's fine to reuse, but otherwise I'd be nervous about it drifting for some reason (primarily if you run `apt install` from within the container I guess)

TomLonergan03, line 111: can you get this and the build directory to be the same? currently running `make` in the dev container errors unless it starts from a blank `build` dir, but it would be nicer if the dev container doesn't need cmake reran after using the build container

TomLonergan03 (changes requested):

docker.sh
DylanCavers: if i change the mount to hyped/$IMAGE_NAME it still fails, not if its a problem with having different containers for the build and dev?

TomLonergan03: if you run `./docker.sh -b` then run `./docker.sh -d`, cd to build, and run `make` it gives the error
```
/bin/sh: 1: cd: can't cd to /home/hyped_build/build
```
I think this is cos the build container mounts the hyped repo to `/home/hyped_build/` whereas dev mounts it to `/home/hyped/`, is it possible to have both in the same place within the container

DylanCavers: for some reason when I change where it's mounted in the run command it doesn't actually change anything in the container which is strange. Will see if I can fix

TomLonergan03 (approved):

licornes-fluos (approved):

PR #51, opened 2023-12-28
Infra - Pi version manager

TomLonergan03:
This will run as a systemd(?) daemon on each pi, and simply waits for a request then carries it out. Currently this does 2 things:
1. `get_hashes` returns a JSON object in the form `{"binary": <hash>, "config": <hash>}`
2. `start` launches the binary on that pi

Config file is a placeholder

Todo:
- [x] Implement start
- [x] Paths on the actual pis
- [x] Set up daemon
- [ ] Package it? Might be useful to have a .deb file
  - Not doing this, but made an install script
- [x] Remove test_client.py


config/pod.toml
davidbeechey, line 1: Is the config going to support multiple pods?

TomLonergan03: It could, by putting each under its own `[heading]`, how much do you care about that?

davidbeechey (approved):
Looks good!

DylanCavers (approved):

config/pod.toml
davidbeechey: The GUI supports multiple pods so it'd be nice, but MQTT broker being moved onto the pod complicates the current multiple pod implementation anyway so I'm happy to just leave it :)

PR #56, opened 2024-01-23
TEL - Fake Data Generation

CThomson2:
In-progress branch. Submitting PR for team lead to review progress made thus far. Documentation attached via readme file.


CThomson2:
# Fake Data Generation Branch


<pre style="color: white">
<h3><strong>Project Structure</strong></h3>
<strong>data-gen</strong>/
| - <a href="#main">main.ts</a> // <em>runs the program</em>
| - <a href="#index">index.ts</a> // <em>handles imports and exports</em>
| - src/
|   | - sensors.ts // <em>self-contained file with classes describing sensor functionality</em>
|   | - sensorData.csv // <em>setup file defining sensor input params with read/write permissions</em>
| - utils/
|   | - data-manager.ts // <em>Singleton class structure used to track, update, upload and store data in real-time</em>
|   | - config.ts // <em>sets up the program and all objects & variables used within</em>
|   | - helpers.ts // <em>contains generic functions used throughout the project</em>
| - tests/ // <em>yet to be implemented</em>
| - tsconfig.json // <em>generic settings for ../server/ directory</em>
| - README.md
</pre>


### Types
New types are defined for this program:
<ul>
    <li><code>LiveMeasurement</code>: extends <code>RangeMeasurement</code>, adding the properties of <code>currentValue</code> and <code>timestep</code>, allowing the adapted use of prewritten data structures.</li>
    <li><code>SensorData</code>: <code>{ [x: string]: LiveMeasurement }</code>. This takes the sensor data and puts them back in the standard <code>Record< string, Measurement></code> format.</li>
    <li><code>StoredData</code>: <code>{ [key: string]: [(number | string)][] }</code>. This object has the sensors/measurements as keys, and at each sensors' timestep an array containing the time stamp and current value are pushed into the sensor's array. While data points will be uploaded in real-time, they are stored for recording and analysis purposes.</li>
    <li><code>InitialState</code>: <code>[key: string]: { dt: number; initialVal: number; }</code>. This interface defines the object which holds the user-specified properties. Again, the key is the sensor name. The user can read and write to a CSV file to specify any or all of the sensors' initial values and the time interval between their generated readings.</li>
</ul>

<span style="display: block; height: 2px; background-color: #bbb;"></span>

## Config
This setup stage imports nested `measurements` object from <code>pods.ts</code> data. The `rangeFilter` function filters this data into an array of only the measurement/sensor objects of type `RangeMeasurement`, and removes duplicate redundancy, then converts it back into an object of similar format to `measurements`. This makes the code readable and familiar. This new object is exported by default.

The array manipulation removes those without a `limits` property (i.e. of type `EnumMeasurement`), and converts all duplicates into equivalent strings to facilitate filtering e.g. 'thermistor_1', 'thermistor_2' -> both become 'thermistor'. Upon the ultimate conversion back into an object (the code uses `Object.fromEntries( Object.entries(measurements) )`), only one 'thermistor' entry is retained due to the JS Object prototype's inherent unique key characteristics.

Then the `readData` helper function is called. This function reads the CSV data, which consists of `[sensor/quantity, time interval, initial value]`. It returns a Promise, which resolves if the resulting object (of type `InitialState`) is not empty. The result is chained to a `then` block, which requires that `unqSensorObj[sensor].currentValue = response[sensor].initialVal`, ensuring the CSV data has the correct sensor names and number of rows.



<span style="display: block; height: 1px; background-color: #bbb;"></span>


## Index
Gathers and exports all exports from relevant project files for ease of access

<span style="display: block; height: 1px; background-color: #bbb;"></span>


## Sensors
Self-contained module containing classes for the different sensor/measurement categories (Navigation, Pressure etc.). Each subclass inherits from the parent Sensor class, reducing repetitive code. The amount of different classes is as small as possible, each combined into groups of similar functionality.

Classes are instantiated once each in `main.ts`. The Sensor class is constructed with a single `LiveMeasurement` sensor object, and its properties are set as `protected readonly`, allowing sub-classes to access them but placing a data barrier from external functions to access or modify these values.

The `_currentValue` variable is mutable, so that only a single instance per class is needed, updating its current value.

Not all sensors/measurements require their own class instance. For instance, in the navigation category only velocity (or the accelerometer reading) needs to be recorded in the class instance. From this value and given a reasonable function of time, the other navigation variables can be calculated accordingly at each time step using basic kinematics and calculus of limts. Additionally, many other measurements have a dependency on pod velocity, generally increasing in proportion to the speed. Temperature is another measurement which dictates certain others, like reservoir pressure for example.

<span style="display: block; height: 1px; background-color: #bbb;"></span>


## Data Manager

<span style="display: block; height: 1px; background-color: #bbb;"></span>


## Main
Main functional file. After importing all necessary objects, classes and functions, it runs the `GenerateDataSeries` iterative function, updating the `DataManager` instance at each iteration.

Imports the adapted `rangeSensors` object from `config.ts` as well as the Sensor class and its sub-classes from `sensors.ts`, which each consist of similar functionality shared by a category of sensors such as Navigation, Pressure etc.

Runs the main loop with user-defined parameters, with the actual functionality and data management in the other files.


<span style="display: block; height: 2px; background-color: #bbb;"></span>

<hr>

Screenshot 2024-01-27 at 10 37 20

davidbeechey:
Moved to #59

PR #58, opened 2024-01-26
SNS - 16-Bit I2C

sa-fx:
Update `II2C` and `HardwareI2C` to support writing data to 16-bit registers on I2C devices
(Required to implement Time of Flight sensor)

- [x] Add overloaded `II2C::writeByteToRegister()` to support `uint16_t` register argument
- [x] Add overloaded `HardwareI2C::writeByteToRegister()` to support `uint16_t` register argument
- [x] Implement private `writeByteToDevice()` function for writing to register addressed by either 8 or 16 bits
- [x] Update surrounding I2C code (`dummy_i2c.hpp\.cpp`, `i2c_commands.hpp\.cpp`)
- [ ] Test code for support with 8-bit register requests (ie, all currently implemented I2C sensors)
- [ ] Test code for support with 16-bit register requests (ie, Time of Flight sensors)

### DIFF
--- a/lib/io/hardware_i2c.cpp
+++ b/lib/io/hardware_i2c.cpp
@@ -70,34 +70,33 @@ core::Result HardwareI2c::writeByteToRegister(const std::uint8_t device_address,
                                               const std::uint8_t register_address,
                                               const std::uint8_t data)
 {
-  const std::uint8_t register_address_array[2] = {0x00, register_address};
-  return writeByteToDevice(device_address, register_address_array, data);
+  const std::vector<std::uint8_t> bytes = {register_address, data};
+  return writeBytesToDevice(device_address, bytes);
 }
 
 core::Result HardwareI2c::writeByteToRegister(const std::uint8_t device_address,
                                               const std::uint16_t register_address,
                                               const std::uint8_t data)
 {
-  // TODOLater - Is this the best way to do this?
-  const std::uint8_t register_address_hi       = register_address >> 8;
-  const std::uint8_t register_address_lo       = static_cast<std::uint8_t>(register_address);
-  const std::uint8_t register_address_array[2] = {register_address_hi, register_address_lo};
-  return writeByteToDevice(device_address, register_address_array, data);
+  const std::uint8_t register_address_hi = register_address >> 8;
+  const std::uint8_t register_address_lo = static_cast<std::uint8_t>(register_address);
+  const std::vector<std::uint8_t> bytes  = {register_address_hi, register_address_lo, data};
+  return writeBytesToDevice(device_address, bytes);
 }
 
 // TODOLater - Test code
-core::Result HardwareI2c::writeByteToDevice(const std::uint8_t device_address,
-                                            const std::uint8_t register_address_array[2],
-                                            const std::uint8_t data)
+core::Result HardwareI2c::writeBytesToDevice(const std::uint8_t device_address,
+                                             const std::vector<std::uint8_t> bytes)
 {
   if (sensor_address_ != device_address) { setSensorAddress(device_address); }
-  const std::uint8_t write_buffer[3] = {register_address_array[1], register_address_array[0], data};
-  const ssize_t num_bytes_written    = write(file_descriptor_, write_buffer, 3);
-  if (num_bytes_written != 3) {
+  const std::size_t number_of_bytes = bytes.size();
+  const std::uint8_t *write_buffer  = bytes.data();
+  const ssize_t num_bytes_written   = write(file_descriptor_, write_buffer, number_of_bytes);
+  if (num_bytes_written != number_of_bytes) {
     logger_.log(core::LogLevel::kFatal, "Failed to write to i2c device");
     return core::Result::kFailure;
   }
-  logger_.log(core::LogLevel::kDebug, "Successfully wrote byte to i2c device register");
+  logger_.log(core::LogLevel::kDebug, "Successfully wrote bytes to i2c device register");
   return core::Result::kSuccess;
 }
 
--- a/lib/io/hardware_i2c.cpp
+++ b/lib/io/hardware_i2c.cpp
@@ -86,7 +86,7 @@ core::Result HardwareI2c::writeByteToRegister(const std::uint8_t device_address,
 
 // TODOLater - Test code
 core::Result HardwareI2c::writeBytesToDevice(const std::uint8_t device_address,
-                                             const std::vector<std::uint8_t> bytes)
+                                             const std::vector<std::uint8_t> &bytes)
 {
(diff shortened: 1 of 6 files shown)


lib/io/hardware_i2c.cpp
```diff
@@ -69,11 +69,31 @@ std::optional<std::uint8_t> HardwareI2c::readByte(const std::uint8_t device_addr
...
+  const std::uint8_t register_address_hi       = register_address >> 8;
+  const std::uint8_t register_address_lo       = static_cast<std::uint8_t>(register_address);
+  const std::uint8_t register_address_array[2] = {register_address_hi, register_address_lo};
+  return writeByteToDevice(device_address, register_address_array, data);
+}
+
+// TODOLater - Test code
+core::Result HardwareI2c::writeByteToDevice(const std::uint8_t device_address,
+                                            const std::uint8_t register_address_array[2],
+                                            const std::uint8_t data)
```
TomLonergan03, line 91: make this more generic, type signature should be 
```suggestion
core::Result HardwareI2c::writeBytesToDevice(const std::uint8_t device_address,
                                            const std::vector<std::uint8t> bytes)
```
and then alter the write to be able to deal with any length of array

lib/io/hardware_i2c.cpp
```diff
@@ -69,11 +69,31 @@ std::optional<std::uint8_t> HardwareI2c::readByte(const std::uint8_t device_addr
 core::Result HardwareI2c::writeByteToRegister(const std::uint8_t device_address,
                                               const std::uint8_t register_address,
                                               const std::uint8_t data)
+{
+  const std::uint8_t register_address_array[2] = {0x00, register_address};
```
TomLonergan03, line 73: this sneakily wouldn't work, it would write `register_address` to register `0x00`

lib/io/hardware_i2c.cpp
```diff
@@ -69,11 +69,31 @@ std::optional<std::uint8_t> HardwareI2c::readByte(const std::uint8_t device_addr
...
+{
+  // TODOLater - Is this the best way to do this?
+  const std::uint8_t register_address_hi       = register_address >> 8;
+  const std::uint8_t register_address_lo       = static_cast<std::uint8_t>(register_address);
+  const std::uint8_t register_address_array[2] = {register_address_hi, register_address_lo};
+  return writeByteToDevice(device_address, register_address_array, data);
+}
+
+// TODOLater - Test code
+core::Result HardwareI2c::writeByteToDevice(const std::uint8_t device_address,
```
TomLonergan03, line 89: `writeBytesToDevice`

lib/io/hardware_i2c.cpp
```diff
@@ -69,11 +69,31 @@ std::optional<std::uint8_t> HardwareI2c::readByte(const std::uint8_t device_addr
...
+{
+  const std::uint8_t register_address_array[2] = {0x00, register_address};
+  return writeByteToDevice(device_address, register_address_array, data);
+}
+
+core::Result HardwareI2c::writeByteToRegister(const std::uint8_t device_address,
+                                              const std::uint16_t register_address,
+                                              const std::uint8_t data)
+{
+  // TODOLater - Is this the best way to do this?
```
TomLonergan03, line 81: this is the best way to do it i'm pretty sure

lib/io/hardware_i2c.hpp
```diff
@@ -28,10 +28,16 @@ class HardwareI2c : public II2c {
   virtual core::Result writeByteToRegister(const std::uint8_t device_address,
                                            const std::uint8_t register_address,
                                            const std::uint8_t data);
+  virtual core::Result writeByteToRegister(const std::uint8_t device_address,
```
TomLonergan03, line 31: as an aside, you were sensors last year do you know why these are virtuals? I don't think we ever inherit (or are likely to?) from this class

sa-fx: Don't know, and I agree: I'm not sure when you would ever inherit from HardwareI2C specifically

lib/io/hardware_i2c.cpp
sa-fx: Seemed slightly hacky, but fair enough

lib/io/hardware_i2c.hpp
```diff
@@ -23,15 +24,20 @@ class HardwareI2c : public II2c {
...
+                                   const std::uint8_t data);
+  core::Result writeByteToRegister(const std::uint8_t device_address,
+                                   const std::uint16_t register_address,
+                                   const std::uint8_t data);
+  core::Result writeByte(const std::uint8_t device_address, const std::uint8_t data);
 
  private:
   void setSensorAddress(const std::uint8_t device_address);
+  core::Result writeBytesToDevice(const std::uint8_t device_address,
+                                  const std::vector<std::uint8_t> bytes);
```
TomLonergan03, line 40: this should be passed as a reference `vector<...> &bytes`

TomLonergan03 (changes requested):
one more thing then I'm happy

sa-fx:
In `HardwareI2c::setSensorAddress()`, should this:

``` 
  sensor_address_ = device_address;
  const int return_value = ioctl(file_descriptor_, I2C_SLAVE, device_address);
  if (return_value < 0) {
    logger_.log(core::LogLevel::kFatal, "Failed to set sensor address");
    return;
  }
```

Not be this:

``` 
  const int return_value = ioctl(file_descriptor_, I2C_SLAVE, device_address);
  if (return_value < 0) {
    logger_.log(core::LogLevel::kFatal, "Failed to set sensor address");
    return;
  }
  sensor_address_ = device_address;
```

Ie, the value of `sensor_address_` is only updated to be `device_address` if `ioctl()` returns no errors

lib/io/hardware_i2c.cpp
sa-fx: If this wouldn't work for the reason I think you're suggesting (that the eventual use of this array in `write()` will take the first element as the register to write to and the second as the value to write to the array), why would this work for the 16-bit addresses used later which are formed in the same way?

If I have a register address `0x0123` that I want to write the value `0x45` to this register, the code as I've written it will send this to `write()` as a buffer `{0x01, 0x23, 0x45}`. How can `write()` differentiate between:
- Attempting to access register `0x01` to write the value `0x4523`
- Attempting to access register `0x0123` to write the value `0x45`

TomLonergan03 (approved):
test it and it's done

lib/io/hardware_i2c.hpp
sa-fx: Update: Ishan said there was a reason for this. @TomLonergan03 What was it, and should these virtuals be reinstated as a result?

licornes-fluos (approved):

PR #59, opened 2024-01-27
TEL - Fake Data Generator

CThomson2:
<hr>

## Final Stage Checklist

1. **Test individual data functionality**
- [x] Motion
- [x] Keyence
- [x] Temperature
- [x] Resistance
- [x] Magnetism
- [x] Levitation
- [ ] Pressure

2. **Test special cases**
- [x] All sensors
- [x] Only dependent sensors
- [x] Different clock frequencies (edge case of unique sampling period for each sensor)

3. **Refactor**
- [x] Remove unneeded documentation and move explanation/overview to readme
- [x] Abstract functionality where suitable
- [x] Assess variable naming and consistency of style

4. OpenMCT implementation with MQTT
- [x] Test program with server

<hr>

# Fake Data Generation Branch

Screenshot 2024-02-05 at 17 23 36

<pre style="color: white">
<h3><strong>Project Structure</strong></h3>
<strong>data-gen</strong>/
| - <a href="#main">main.ts</a> // <em>runs the program</em>
| - <a href="#index">index.ts</a> // <em>handles imports and exports</em>
| - <a href="#types">types.ts</a> // <em>defines complex types used throughout program</em>
| - src/
|   | - sensors.ts // <em>self-contained file with class definitions for each sensor type</em>
|   | - dataManager.ts // <em>SensorMap class to oversee data generation and abstract base class Sensor</em>
|   | - sensorUtils.ts // <em>class of self-contained mathematical utility helper methods for data generation</em>
| - env/
|   | - config.ts // <em>Sets up data objects for use during runtime</em>
| - tests/ // <em>yet to be implemented</em>
| - tsconfig.json // <em>generic settings for ../server/ directory</em>
| - README.md
</pre>

<span style="display: block; height: 4px; background-color: #555;"></span>

## Data Generation Logic

There are 7 types of sensors/measurements. Their categories are listed below with indents representing hierarchy of dependencies. The parenthesis indicate variables their `nextReadings` methods depend upon.

- | - Motion (_time_)
- | - Keyence (_displacement_)
- | - Temperature (_velocity_)
- | - Resistance (_temperature_)
- | - Pressure (_velocity & temperature_)
- | - Electromagnetism (_velocity_)
- | - Levitation (_hall effect for calibration_)

1. **Motion** (accelerometers) are calculated first at every timestep. They provide readings for acceleration, velocity and displacement, which are saved for that instant and uploaded to the server. Their values are estimated based on an assumed logistic function for velocity over time, which is the analytical function the velocity would follow in order to reach max speed as quickly as possible. The function parameters are defined such that derivative, acceleration, remains within critical limits.
2. **Keyence** optical sensor counts stripes/poles, so it is directly proportionate to displacement. By matching the track length to the corresponding number of poles, and rounding the value down to get an integer, keyence values can be calculated trivially. However as there a few sensors on board, and assuming they are spaced equally, additional logic can be implemented to calculate each sensor's offset from the pod's nose (datum for displacement reading) and factor that in. This is the only sensor without added noise, as it is an integer. Instead noise could added in the form of a ± X mm tolerance.
3. **Temperature** sensing thermistors are calculated to be proportional to pod velocity, as this induces greater air friction and heat emanating from internal components. The temperature reading published is the average of the thermistor values, which all have individual additional noise.
4. **Resistance** sensors generate a near-constant value over the pod run, with a slight proportional increase with temperature of around 5*10^(-3) Ohm/K (material's temperature coefficient of resistance).
6. **Pressure** gauges have different functions across the pod. The push/pull sensors are calculated assuming no other air infiltration other than the external pneumatic tubing. After subtracting an internal pressure drop, the four pressure gauge readings are found by maintaining a constant airflow, so sum(P_pull) = sum(P_push) - losses.
7. **Electromagnetism** is the class name for the hall effect sensors controlling monitoring the magnetic field inducing levitation. The magnetic field force is controlled by operator and fixed at a setpoint value corresponding to the desired levitation gap height. When the pod reaches a threshold velocity, the electromagnets are turned on the hall effect sensors register an instantaneous step up to the setpoint value. Interference comes in the form of the magnetic field arising from the LIM, so as velocity changes rapidly the readings fluctuate sharply.
8. **Levitation** is measured by the time-of-flight sensors. When the electromagnets turn on, it follows an exponential increase to its setpoint, following the balance of forces of gravity and the inverse square law of magnetic repulsion/attraction. The reading then oscillates around the setpoint and decays quickly to steady state.

All of the functions described require numerical parameters which are currently hardcoded in a file in fake/env/ and later could be set from the GUI.

<span style="display: block; height: 4px; background-color: #555;"></span>

## Types
New types are defined for this program:
<ul>
    <li>
    <code>Readings = Record</code><<code>string, number</code>>
    Provides an object type to store live sensor readings during runtime, e.g. 'accelerometer_3: 2.3'.
    </li>
     <li>
     <code>ReadingsMap = Record</code><<code>string, Readings</code>>
     Object type to store the readings for all sensors, categorised by keys representing sensor types.
     </li>
         <li>
     <code>RunData = ReadingsMap[]</code>
     Array of all readings. New data is pushed at each timestep, where the array indices represent timesteps. While data points will be uploaded in real-time, they are stored for recording and analysis purposes.
     </li>
<br>
    <li><code>LiveMeasurement = RangeMeasurement & { quantity: number; readings: Readings }</code>
     Integrates fixed sensor data with dynamic data along with the quantity of sensors of a specific 
     type e.g. pressure gauges. For the variables representing physical sensors, not derived measurements.
     <li><code>SensorData = Record</code><<code>string, LiveMeasurement</code>>
This takes the sensor data and puts them back in the standard <code>Record</code><<code>string, Measurement></code> format.
    </li>
<br>
     <li>
     <code>SensorInstance</code><<code>T extends new (...args: any[]) => any</code>><code> = InstanceType</code><<code>T</code>></code><br>
     Type used to create sensor instances. Generic parameter T example: <code>SensorInstance</code><<code>typeof sensors[keyof typeof sensors]</code>>, where sensors is an object of individual sensor classes.
     </li>
</ul>

### Type Relations (with example data)

 * Readings: `{ acceleration: 2, displacement: 5.3, ... }`
 
 * ReadingsMap: `{ motion: Readings } & { pressure: Readings } & { ... } & ...`
 = `{ motion: { acceleration: 2, displacement: 5.3, ... }, pressure: { ... }, ... }`

 * RunData: `[ { motion: { acceleration: 2, displacement: 5.3, ... }, pressure: { ... }, ... }, { ... }, ... ]`
 * `~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~`
 * RangeMeasurement: `{ name: 'acceleration', key: 'acceleration', unit: 'm/s^2', type: 'motion', format: 'float', limits: { warning: { low: 0, high: 0 }, critical: { low: -10, high: 10 } }, rms_noise: 0.1, sampling_time: 0.1 }`

 * LiveMeasurement: `{ quantity: 2, readings: Readings & {...RangeMeasurement} }`
 = `{ quantity: 2, readings: { acceleration: 2, displacement: 3 } & {...RangeMeasurement} }`
 
 * SensorData: `{ motion: { quantity: 2, readings: { acceleration: 2, displacement: 5.3, ... } }, pressure: { ... }, ... }, { ... }, ... }`

<span style="display: block; height: 2px; background-color: #bbb;"></span>

## Config

This setup stage configures and exports two data objects to be used throughout the rest of the program. First, the pod's sensor measurements data is imported from <code>pods.ts</code>. The object is filtered to result in an array of only the measurement/sensor objects of type `RangeMeasurement` (i.e. those with numerical readings suitable for simulating data). The resulting object has the same shape as the measurements object, maintaining consistency and familiarity throughout the codebase.

The rest of file initialises and fills the new sensorData object (type SensorData). This object categorises all similar sensors of the same type (defined by the physical quantity(ies) they measure, e.g. motion variables), setting the keys to the sensor type and adding data on all constituents to the new readings property.

Initial values are required for `sensorData[sensor].readings`, and the program hard codes appropriate initial conditions for each sensor. Motion sensors (incl. keyence), levitation height and hall effect sensors begin at zero, for obvious reasons. External pressure monitors are initialised to 1 bar, while internal reservoir pressures are assumed to be near the middle of their critical limiting ranges (although this could be closer the lower end to give leeway for pressure changes, as reservoir pressure will likely only rise as the pod speeds up).

Helper functions are defined to set initial conditions, as well as count the amount of sensors of a given type, which generate values for the quantity and readings properties for each sensor type in sensorData.

The file's object manipulation firstly removes sensors without a `limits` property (i.e. of type `EnumMeasurement`) to create `measurements`, then converts all duplicate sensor names into equivalent strings to facilitate the categorisation and remove redundancy, e.g. 'thermistor_1', 'thermistor_2' -> both become 'thermistor'. Upon the ultimate conversion back into an object, only one 'thermistor' entry would be retained due to the JS Object prototype's inherent unique key characteristics.

<span style="display: block; height: 1px; background-color: #bbb;"></span>

## Index
Gathers and exports all exports from relevant project files for ease of access

<span style="display: block; height: 1px; background-color: #bbb;"></span>

## Sensors

Self-contained module containing classes for the different categories: _Physical quantity (sensor)_

- Navigation (accelerometers)
- Pressure (various gauges)
- Temperature (thermistors)
- Electromagnetic force (hall effect sensors)
- Optical (keyence sensors)
- Resistance (multimeters)
- Levitation (Time-of-Flight sensors)

Each subclass inherits from the abstract Sensor class, reducing repetitive code as their logic all shares the same purpose - generate data, update dynamic class variables in Sensor & SensorMap, and upload to MQTT. The abstract class defines an abstract update method, which sees a different implementation across the different sensors. It also has a defined update method, which is the same for all sensors.

The sensor instances are not independent, but - at least for a realistic simulation of transient data - rely upon the readings taken from other sensors. Motion sensors (i.e. accelerometers) are at the top level of this hierarchy, as they define parameters required to calculate all other measurements, either directly or otherwise (the latter exemplified in the case of the pressure readings which depend on temperature, which increases with velocity).

While some sensor logic is fairly straightforward (keyence, power line resistance, temperature), others are a little more complex and interdependent. Looking at the motion category, a function has to be chosen to describe the dynamics. Velocity is the obvious candidate, as the high-level goal of a successful pod run is to bring the pod to top speed as quickly as possible while remaining within operational limits. Therefore velocity is the variable which can intuitively be described by a function. The logistic curve was chosen, with parameters such that acceleration peaks at 5m/s<sup>2</sup>. From this analytical function, the accelerometer readings are generated by taking the derivative at a given time and adding noise to each sensor. The three output variables: acceleration, velocity and displacement, are calculated by averaging sensor readings for acceleration and using discrete calculus / basic kinematics for the other two.

After the values have been generated, they are uploaded to the server.
 
Other sensors have their own logic, with the caveat that many assumptions are made. However this is not an issue, regarding the purposes of this fake data generation program. A notable mention is Levitation and Electromagnetism. It is assumed that the levitation systems switch on once the pod has reached a specified threshold velocity. Therefore both the hall effect and time-of-flight sensors register zero until then. When that velocity is reached, power is turned on to the electromagnets inducing levitation. The gap height between the pod and track follows a similar logistic curve to velocity, with the addition of an exponentially decaying sinusoidal pattern after shooting up to its setpoint gap height, before gap height reaches a steady state.

More complexity could be introduced by taking into account fluctuations in the magnetic field measured due to changes in the magnetic force from the propulsion system. However, as the input/control variable of current is operator-controlled and has a matching setpoint to the gap height, this interference issue could be avoided by fixing the levitation gap height in place once it has reached its expected steady state, essentially decoupling it from following the hall effect sensor readings, until the time for the pod to slow down and rest on the track again comes. However, the time-of-flight sensors will register random noise throughout operation. The issue this poses is that if the gap height actually veers off course, the current supplied must be adjusted to bring it back to its setpoint. So noise cannot be ignored completely in the calculations, which is where the exponential moving average plays a role. By averaging the last N time-of-flight readings - weighting the recent ones more for quicker responsiveness - outliers can be spotted quickly while random fluctuating noise is ignored.

Classes are instantiated once each in `main.ts`. The Sensor class is constructed with a single `LiveMeasurement` sensor object, and its properties are set as `protected readonly`, allowing sub-classes to access them but placing a virtual barrier preventing external code to access or modify these values.

The `readings` object is mutable, so only a single instance per class is needed, continually self-updating.

Dividing the many pod measurements into a few select classes is achievable due to the nature of the relationship between a sensor reading and the related variables that can be derived from it, such as displacement from the accelerometer readings.

<span style="display: block; height: 1px; background-color: #bbb;"></span>

## Sensor Utilities

Sensors classes generally only have their update methods, while outsourcing functional complexity to the Utilities class, which provides self-contained static methods as pure functions which aid in generating data. The utilities provided consist of data simulation estimated functions, such as the logistic curve for velocity over time, other mathematical functions, along with methods used to generate random noise and find moving averages.

<span style="display: block; height: 1px; background-color: #bbb;"></span>

## Data Manager

Holds the SensorMap class. This class manages the asynchronous and sensor-independent sampling frequencies, runs the corresponding sensors at each time step, updates global dynamic data, and sets a delay at each iteration lasting the shortest time period between the current time and next time at which a sensor is due to be sampled. It corresponds with the Sensor class and its individual instances.

<span style="display: block; height: 1px; background-color: #bbb;"></span>

## Main

Main functional file. After importing all necessary objects, classes and functions, it runs the `generateDataSeries` iterative function, outsourcing all the sensor logic and uploading responsibility to the sensor classes at each iteration.

Takes in CLI input:
- The `--randomise` flag, set to false by default. Setting this flag true induces the code to completely randomise each readings. Used for debugging purposes and validating the program's output.
- The `--runtime [time in seconds]` flag with a time parameter. This defines the real time the program will run for. The main loop requires that `Data.now() - startTime` be less than than `runTime` to continue iterating.

<span style="display: block; height: 2px; background-color: #bbb;"></span>

<hr>

Screenshot 2024-01-27 at 10 37 20


CThomson2:
- Removed initial_value property from pods object, as it's not an inherent property, and would reduce readability during runtime if each step refers to the current value as "initial_value". Adjusted pod.types types to reflect this.
- For the items in the measurements object which are physical sensors, I identified their expected noise (RMS), to use as the std dev. for a normal distribution random number generator. Set noise values for derived measurements to 0, as their random noise will be propagated down during runtime from their respective sensor reading.
- Categorised all measurements by their sensor type (this was only the navigation items), and replaced 'type' property for the sensor type, leaving the 'key' to effectively be each item's individual ID.
- Added a sampling_time property. While this isn't a fixed value, as clock frequency is adjustable, it will be constant during each run, unlike the current_value property.

telemetry/packages/server/fake/index.ts
```diff
@@ -10,5 +10,30 @@ export { pods } from "../../constants/src/pods/pods";
...
 
 export { measurements, sensorData } from "./env/config"
-export { SensorMap } from './src/sensorManager';
-export { Utilities } from './src/sensorUtils';
\ No newline at end of file
+export { SensorMap } from './src/dataManager';
+
+export { Utilities } from './src/sensorUtils';
+
+// Sensor classes should be encased in an array so as to maintain correct ordering
```
CThomson2, line 17: Did this in dataManager, ignore commented code here

CThomson2:
Just realised it can be simplified, no need for a runTime parameter we just use system.exit once displacement reaches the end of the road using its 100m limit

telemetry/packages/fake/src/sensorClasses/sensorIndex.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Motion } from "./motion";
+import { Keyence } from "./keyence";
+import { Pressure } from "./pressure";
+import { Temperature } from "./temperature";
+import { Resistance } from "./resistance";
+import { Magnetism } from "./magnetism";
+import { Levitation } from "./levitation";
+
+// Package all sensors into one object to be iterated through easily in SensorManager
+interface sensorObj<T extends Sensor> {
```
davidbeechey, line 12: Since this type isn't reused anywhere else, you're probably best annotating `sensors` below with `Record<string, typeof Sensor>` which would enforce the correct typing of that object

telemetry/packages/fake/src/sensorClasses/sensorIndex.ts
davidbeechey, line 1: rename to `index.ts`

telemetry/packages/fake/src/baseSensor.ts
davidbeechey, line 1: remove unused class variables

davidbeechey, line 1: rename to `base.ts`?

telemetry/packages/fake/src/config.ts
```diff
@@ -0,0 +1,117 @@
...
+      // Normalise sensor name to its category's base name, as this function is called only once for each sensor type
+      // replace name with base name, effectively categoring sensors by type and function
+      acc[key].name = key.replace(/_[^_]*\d$/, '');
+    }),
+    acc
+  ),
+  {} as Record<string, RangeMeasurement>,
+);
+
+// console.log(measurements)
```
davidbeechey, line 20:
```suggestion
```

telemetry/packages/fake/src/config.ts
```diff
@@ -0,0 +1,117 @@
...
+              (name) =>
+                !name.endsWith('avg') && measurements[name].type == data.type,
+            )
+            .map((el) => [el, getInitialValue(measurements[el])]),
+        ),
+      } as LiveReading,
+    ]),
+);
+
+// console.log(sensorData);
```
davidbeechey, line 114:
```suggestion
```

telemetry/packages/fake/src/dataManager.ts
davidbeechey, line 1: Filename is `dataManager` but the class is named `SensorManager` - best to rename one so that they're the same

davidbeechey, line 1: Quite a lot of unnecessary comments in this file, best to remove those which are described e.g. by variable names

telemetry/packages/fake/src/index.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * Main file which initialises the generation of the data series and uploads to GUI in real time.
+ * @param runTime (CLI) simulation time in ms (not real time, based on sensor timesteps)
```
davidbeechey, line 3: Doesn't look like this CLI argument is ever used? I assume it was replaced with the `trackLength` limit instead?

telemetry/packages/fake/src/index.ts
davidbeechey, line 1: Remove commented-out code from this file...

telemetry/packages/fake/src/logicTests/motionTest.js
davidbeechey, line 1: Stray JS file from `tsc` output - delete

telemetry/packages/fake/src/logicTests/motionTest.ts
davidbeechey, line 1: If you're planning to keep this test, probably rename this folder `tests` and this `motion.ts`

davidbeechey (changes requested):
Hi Conrad, thanks for your great work on this! Let me know if you have any questions :)

telemetry/packages/fake/fig/electromagnetic_force.png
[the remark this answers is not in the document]
TomLonergan03: what is it used for? if it's needed in the wiki add it to the wiki repo instead, if it's needed in a readme it can stay (but I would encourage README stuff to go in the wiki instead generally)

telemetry/packages/fake/src/base.ts
```diff
@@ -0,0 +1,71 @@
...
+   * Returns the Readings object, filtered into the values to be published with MQTT
+   * For motion, only acceleration, velocitity and displacement are uploaded
+   * Accelerometers are used to estimate readings, then these values are
+   *   propagated to the three variables above
+   * @param t time in seconds
+   */
+  abstract getData(t: number): Readings;
+
+  getRandomData(readings: Readings): Readings {
+    // console.log(`Running randomData...`);
```
davidbeechey, line 61:
```suggestion
```

telemetry/packages/fake/src/base.ts
```diff
@@ -0,0 +1,71 @@
...
+  }: LiveReading) {
+    Object.assign(this, {
+      type,
+      format,
+      limits,
+      rms_noise,
+    });
+    this.delta_t = sampling_time / 1000; // convert ms to s
+    this.time = 0;
+    // console.log('/nInitial values verification:', readings);
```
davidbeechey, line 45:
```suggestion
```

telemetry/packages/fake/src/base.ts
davidbeechey, line 1: should this be in the sensors folder?

telemetry/packages/fake/src/sensorClasses/pressure.ts
```diff
@@ -0,0 +1,91 @@
...
+  private coefficients = {
+    lossFactor: 0.05, // 5% internal pressure losses
+    stagnation: 0.5,
+    wake: 0.25,
+    brakingFactor: 0.05, // to calculate pressure increase due to braking force
+  };
+
+  constructor(data: LiveReading) {
+    super(data);
+    // this.prevVals = Array<number>(10).fill(0);
```
davidbeechey, line 23:
```suggestion
```

telemetry/packages/fake/src/sensorClasses/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  | typeof Magnetism
+  | typeof Levitation;
+
+// Package all sensors into one object to be iterated through easily in SensorManager
+export type sensorObj = {
+  [sensor: string]: SensorType;
+};
+
+// Instance type for sensor classes
+export type SensorInstance<T extends new (...args: any[]) => any> =
```
davidbeechey, line 25: this is a weird type def...

telemetry/packages/fake/src/base.ts
```diff
@@ -0,0 +1,71 @@
...
+  public static isSampled: Record<string, boolean> = {};
+  // Stores most recent sensor readings for all sensors, accessible by all sensors
+  // Null is used to indicate that the sensor has not been sampled at the current time
+  public static lastReadings: Record<string, Readings> = {};
+
+  // Sensor properties
+  readonly type: string; // sensor type (same as the name of the object in sensorData)
+  // readonly key: string; // sensor id, but this not as a top-level property, instead the readings all have unique keys
+  readonly format: 'float' | 'integer'; // for random ternary logic (keyence is integer, rest are float)
+  // readonly unit: string; // if we want to display units on UI, probably not necessary as front end will already do this
```
davidbeechey, line 20:
```suggestion
```

telemetry/packages/fake/src/base.ts
```diff
@@ -0,0 +1,71 @@
...
+  public static nextSamplingTimes: Record<string, number>;
+  // Records whether each sensor has been sampled at the current time with a boolean flag for each
+  public static isSampled: Record<string, boolean> = {};
+  // Stores most recent sensor readings for all sensors, accessible by all sensors
+  // Null is used to indicate that the sensor has not been sampled at the current time
+  public static lastReadings: Record<string, Readings> = {};
+
+  // Sensor properties
+  readonly type: string; // sensor type (same as the name of the object in sensorData)
+  // readonly key: string; // sensor id, but this not as a top-level property, instead the readings all have unique keys
```
davidbeechey, line 18:
```suggestion
```

davidbeechey (changes requested):
Just a few minor things to fix. Also rename `sensorClasses` to just `sensors` (bit cleaner).

telemetry/packages/fake/src/sensorClasses/index.ts
CThomson2: We decided to leave it for now, necessary for type recognition. The type gives a structure to items in the array of sensor instances, so TS recognises that .getData, .getRandomData and the shared static attributes are defined on any given sensor instance.

telemetry/packages/fake/src/sensorClasses/sensorIndex.ts
CThomson2: Removed the redundant type def, although it has to be typeof Motion | ... | Levitation as they have different structure.

davidbeechey (approved):
Looks good, well done!! 🎉

TomLonergan03 (approved):
LGTM, lfg

PR #60, opened 2024-01-27
SNS - Optical Flow Sensor (look at this one)

licornes-fluos:


lib/sensors/optical_flow.cpp
```diff
@@ -0,0 +1,58 @@
...
+// (https://github.com/bitcraze/Bitcraze_PMW3901/blob/master/src/Bitcraze_PMW3901.cpp)
+std::uint8_t OpticalFlow::getDeltaX(std::shared_ptr<io::ISpi> spi)
+{
+  std::uint8_t *x_low;
+  std::uint8_t *x_high;
+
+  spi->read(kXLowAddress, x_low, 1);
+  spi->read(kXHighAddress, x_high, 1);
+
+  return static_cast<std::int16_t>((*x_high << 8) | *x_low);
```
TomLonergan03, line 46: check that this actually works, it might shift `x_high` by 8 which 0s it out before the cast applies

lib/sensors/optical_flow.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <optional>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/spi.hpp>
+
+namespace hyped::sensors {
+
+class OpticalFlow {
+  // include "magic sauce" optimisation?
```
TomLonergan03, line 15: these kinds of comments should be `TODOLater`s

TomLonergan03 (changes requested):

PR #67, opened 2024-02-01
MTC - CAN Messages

jpfbastos:


lib/motors/can_messages.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+core::Result CanMessages::CanError(int error)
+{
+  io::CanFrame frame;
+
+  frame.can_id  = convertToBytes(error, 1)[0];
+  frame.can_dlc = 8;
+  frame.data[0] = 0x00;
+  frame.data[1] = convertToBytes(/*location*/ 0, 2)[0];  // wherever the location is
+  frame.data[2] = convertToBytes(0, 2)[0];
```
TomLonergan03, line 42: i don't quite get what's going on here. I think you want to call `convertToBytes(error, 2)` or something then fill `data[0]..[1]` with the indexes of that?

or what do you mean by wherever the location is

lib/motors/can_messages.hpp
```diff
@@ -0,0 +1,26 @@
...
+class CanMessages {
+  enum Operations { kRead, kWrite };
+  enum Locations { temperature, accelerometer };
+
+ public:
+  core::Result CanSend(int operation, int location, int data);
+  core::Result CanError(int error);
+
+ private:
+  CanMessages(std::shared_ptr<io::ICan> can);
```
TomLonergan03, line 20: the constructor should be public, and define the implementation for it

lib/motors/can_messages.cpp
jpfbastos: Oh that was just because since I assumed all messages go to a central computer, the value could be hardcoded as a constant, and so it did not require a variable to handle it - i should have specified that in the code sorry

lib/motors/can_messages.cpp
```diff
@@ -0,0 +1,64 @@
+#include "can_messages.hpp"
+
+namespace hyped::motors {
+
+CanMessages::CanMessages(std::shared_ptr<io::ICan> can) : can_(can)
```
TomLonergan03, line 5: very nitpicky, but `can_(std::move(can))` is supposedly more efficient and has been enforced so far

lib/motors/can_messages.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+}
+
+std::vector<std::uint8_t> CanMessages::convertToBytes(std::uint64_t value, std::size_t length)
+{
+  std::vector<std::uint8_t> bytes = {};
+
+  for (int i = 0; i < sizeof(int); i++) {
```
TomLonergan03, line 13: this should be should be `size_t`, and `i < length`

lib/motors/can_messages.cpp
```diff
@@ -0,0 +1,64 @@
...
+    std::uint8_t byte = (value >> (8 * (length - 1 - i))) & 0xFF;
```
TomLonergan03, line 14: `const`

lib/motors/can_messages.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <core/types.hpp>
+#include <io/can.hpp>
+
+namespace hyped::motors {
+
+class CanMessages {
+  enum Operation { kRead, kWrite };
+  enum Location { kTemperature, kAccelerometer };
+  enum Error { Error1, Error2 };
+  static constexpr std::uint8_t kErrorId = 0;  // TBD
```
TomLonergan03, line 15: `TODOLater decide`  as the comment

lib/motors/can_messages.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <memory>
+#include <vector>
+
+#include <core/types.hpp>
+#include <io/can.hpp>
+
+namespace hyped::motors {
+
+class CanMessages {
```
TomLonergan03, line 11: can you come up with a better name for the class?

TomLonergan03 (changes requested):
Last things

lib/motors/can_messages.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+class VectorControlCanMessages {
+  enum Operation { kRead, kWrite };
+  enum Location { kTemperature, kAccelerometer };
+  enum Error { Error1, Error2 };
+  static constexpr std::uint8_t kErrorId = 0;  // TODOLater decide
+
+ public:
+  VectorControlCanMessages(std::shared_ptr<io::ICan> can);
+  core::Result canSend(Operation operation, Location location, std::uint64_t data);
```
TomLonergan03, line 19: call this `sendResponse`

lib/motors/can_messages.hpp
```diff
@@ -0,0 +1,27 @@
...
+  core::Result canError(Error error);
```
TomLonergan03, line 20: `sendError`

lib/motors/can_messages.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <memory>
+#include <vector>
+
+#include <core/types.hpp>
+#include <io/can.hpp>
+
+namespace hyped::motors {
+
+class VectorControlCanMessages {
```
TomLonergan03, line 11: `ControllerCanProcessor`, and implement `ICanProcessor` for this class. Currently, when it receives a message have it respond with an `Error::kInvalidOperation` error message. (this is an increase in scope for the PR, but might as well do it now)

TomLonergan03 (changes requested):
lmk if you have questions about implementing `ICanProcessor`

lib/motors/can_messages.cpp
```diff
@@ -0,0 +1,72 @@
...
+  frame.data[6] = 0x00;
+  frame.data[7] = 0x00;
+
+  const core::Result result = can_->send(frame);
+  return result;
+}
+
+core::Result ControllerCanProcessor::receiveMessage(io::CanFrame frame)
+{
+  const core::Result result = io::ICanProcessor::processMessage(frame);
```
TomLonergan03, line 67: you don't need to call this

lib/motors/can_messages.cpp
```diff
@@ -0,0 +1,72 @@
...
+  sendError(Error::kInvalidOperation);
```
TomLonergan03, line 68: use the result from this as the return

TomLonergan03 (changes requested):
pretty much exactly what I was thinking, two changes

lib/motors/can_messages.cpp
jpfbastos: so would the function just be like this?

`const core::Result result = sendError(Error::kInvalidOperation);`
`return result;`

TomLonergan03: yes

TomLonergan03 (approved):
looks good, good job!

lsalhani (approved):
lgtm

PR #70, opened 2024-02-12
Infra - Add CMake support for STM32

TomLonergan03:
This is very very WIP - currently only builds by making a build dir in `src/stm32_test` and running cmake from there

TODO
- [x] Cut out everything unnecessary
- [x] Combine with current build system
- [ ] Make nice


TomLonergan03:
This is not gonna happen this year :(

PR #71, opened 2024-02-12
Vector Control General Functionality

ImagineWagonsInSky:
Class for Vector Control, header and main file. This is the initial overview of how vector control will work. We just need it now to merge with CAN messages.

### DIFF
--- a/lib/motors/vector_controller.cpp
+++ b/lib/motors/vector_controller.cpp
@@ -1,0 +1,94 @@
+#include <iostream>
+#include <cmath>
+
+#include "vector_controller.hpp"
+
+namespace hyped :: motors {
+    // TODO steps for vector control
+    // TODO defining class most likely in header file
+    // TODO need to look into how the scope is used in HYPED (using scope resolution operator)
+    VectorController::VectorController() {
+        Kp = 1.0;  
+        Ki = 0.1; 
+        IdRef = 0.0;  
+        IqRef = 0.0;  
+        theta = 0.0; 
+    }
+
+            // Step 1: Measure stator currents and rotor velocity
+            void VectorController::measureCurrentsAndVelocity(core::Float ia, core::Float ib, core::Float ic, core::Float velocity) {
+                Ia = ia;
+                Ib = ib;
+                Ic = ic;
+                rotorVelocity = velocity;
+            }
+
+            // Step 2: Convert to 2-axis system
+            void VectorController::convertTo2Axis() {
+                // Perform transformation from 3-phase currents (Ia, Ib, Ic) to 2-axis (Id, Iq)
+                // Implement the Clarke transformation
+                core::Float Ialpha = Ia;
+                core::Float Ibeta = (Ia + 2*Ib)/sqrt(3);
+                
+                // Step 3: Rotate coordinate system to align with rotor flux
+                // Calculate the transformation angle based on the rotor flux alignment
+                // Implement the Park transformation
+                Id = cos(theta) * Ialpha + sin(theta) * Ibeta;
+                Iq = -cos(theta) * Ialpha + sin(theta) * Ibeta;    
+            }
+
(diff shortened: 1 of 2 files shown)


lib/motors/vector_controller.cpp
```diff
@@ -0,0 +1,94 @@
...
+                // rotor magnetizing flux
+                core::Float integral_term_Id = Ki * IdError;
+                Vd = Kp * IdError + integral_term_Id;
+
+                // torque output of motor
+                core::Float integral_term_Iq = Ki * IqError;
+                Vq = Kp * IqError + integral_term_Iq;
+            }
+
+            void VectorController::parkTransform() {
```
lsalhani, line 68: This should be inverse park

lsalhani (commented):
Quick look

lib/motors/vector_controller.cpp
ImagineWagonsInSky: Sorry about that. Changes done in second commit.

ImagineWagonsInSky:
Probably don't even need to review. We are apparently changing to scalar control now.

lib/motors/vector_controller.hpp
```diff
@@ -0,0 +1,33 @@
...
+  core::Float rotorVelocity;
+  core::Float Id;
+  core::Float Iq;
+  core::Float Vd;
+  core::Float Vq;
+  core::Float IdRef;
+  core::Float IqRef;
+  core::Float Kp;
+  core::Float Ki;
+  core::Float theta;
```
TomLonergan03, line 20: 1. Rename most of these to be clear on what they actually are
2. Variables should be named in `lower_snake_case` and class data members should have a trailing `_`, so `current_phase_a_` etc.
3. Move the private block after public.

lib/motors/vector_controller.cpp
```diff
@@ -0,0 +1,92 @@
...
+            void VectorController::inverseParkTransform() {
+                core::Float Valpha = Vd * cos(theta) - Vq * sin(theta);
+                core::Float Vbeta = Vd * sin(theta) + Vq * cos(theta);
+
+                Vd = Valpha;
+                Vq = Vbeta;
+            }
+        
+
+        int main() {
```
TomLonergan03, line 75: get rid of the main function

lib/motors/vector_controller.cpp
```diff
@@ -0,0 +1,92 @@
...
+            // Measure stator currents and rotor velocity
+            void VectorController::measureCurrentsAndVelocity(core::Float ia, core::Float ib, core::Float ic, core::Float velocity) {
+                Ia = ia;
+                Ib = ib;
+                Ic = ic;
+                rotorVelocity = velocity;
+            }
+
+            // Convert to 2-axis system
+            void VectorController::convertTo2Axis() {
```
TomLonergan03, line 25: these functions should take parameters and return values instead of doing random stuff to internal variables, cos until the end of this function there will be a `Id` and `Iq` which are outdated

lib/motors/vector_controller.hpp
```diff
@@ -0,0 +1,33 @@
...
+ public:
+  VectorController();
+
+  ~VectorController();
+
+  void measureCurrentsAndVelocity(core::Float ia, core::Float ib, core::Float ic, core::Float rotorVelocity);
+  void convertTo2Axis();
+  void inverse2Axis();
+  void piControl();
+  void inverseParkTransform();
```
TomLonergan03, line 31: most of these should be private, and used by public methods

Pass in the new currents and velocity and do the rest of it internally unless these methods definitely need to be called elsewhere

TomLonergan03 (changes requested):
Good start, run clang-format over everything. Also, what is used as the actual PWM duty cycle? Or what values will that be calculated from?

PR #72, opened 2024-02-12
SNS - Brakes sensor

tahn1234:


lib/sensors/brakes.hpp
```diff
@@ -0,0 +1,39 @@
...
+        std::uint8_t off; // if distance is further, it is off. 2 and 5
+        std::shared_ptr <io::IGpioReader> gpio_reader_;
+        core::ILogger &logger_;
+    };
+}
+
+//#ifndef HYPED_2024_BRAKES_H
+//#define HYPED_2024_BRAKES_H
+//
+//#endif  // HYPED_2024_BRAKES_H
```
licornes-fluos, line 39: From Tom on old PR:
you don't need this stuff, #pragma once does the same thing in a nicer way

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,37 @@
+#include "brakes.hpp"
+
+namespace hyped::sensors {
+
+std::optional<Brakes> Brakes::create(core::ILogger &logger,
+                                     std::shared_ptr<io::IGpio> gpio,
+                                     const std::uint8_t new_pin)
```
licornes-fluos, line 7: From Tom on old PR: 
call this pin

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,37 @@
...
+{
+  const auto reader = gpio->getReader(new_pin, io::Edge::kNone);
```
licornes-fluos, line 9: From Tom:
getWriter right? or am I misunderstanding and this PR is for monitoring if the brakes are deployed?

licornes-fluos:
Make modifications so that it build and push.
You need to clang-format your two files. For this, go to the folder where they are and in the command line write `clang-format -i nameOfFile`.

lib/sensors/brakes.cpp
tahn1234: @TomLonergan03 Can you explain this? I am confused why it has to be getWriter not getReader

TomLonergan03: What are you trying to do? Is this what deploys the brakes or is it a sensor that checks if the brakes are deployed? If it's deploying the brakes then it sends a gpio signal, so you need `getWriter`, otherwise you're right

licornes-fluos: We're determining if the brakes are on or not based on distances. So the code is fine.

cmake-build-debug/CMakeFiles/clion-Debug-log.txt
TomLonergan03, line 1: delete this file

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+}
+
+Brakes::~Brakes()
+{
+}
+
+bool Brakes::highLow()
+{
+  auto sig = gpio_reader_->read();
```
TomLonergan03, line 32: `const`, call this `optional_reading`

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+  core::DigitalSignal signal = sig.value();
```
TomLonergan03, line 34: check if this exists, i.e.
```c++
if (!optional_reading) {
    return std::nullopt;
}
const auto reading = *optional_reading;
```

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,38 @@
...
+      on(),
+      off()
+{
+}
+
+Brakes::~Brakes()
+{
+}
+
+bool Brakes::highLow()
```
TomLonergan03, line 30: rename this function to `isClamped` or something

lib/sensors/brakes.cpp
TomLonergan03: this also will need changing the return type to `std::optional<bool>`

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+  auto sig = gpio_reader_->read();
+
+  core::DigitalSignal signal = sig.value();
+  if (signal <= core::DigitalSignal::kHigh) { return true; }
```
TomLonergan03, line 35: i'm pretty sure this is always true cos `DigitalSignal::kLow` is represented by 0

lib/sensors/brakes.hpp
```diff
@@ -0,0 +1,32 @@
...
+  //  bool open();
+  bool highLow();
+
+ private:
+  Brakes(core::ILogger &logger, std::shared_ptr<io::IGpioReader> gpio_reader);
+
+ private:
+  std::uint8_t pin_;
+  std::uint8_t on;   // if distance is closer, it is on 0 and 2
+  std::uint8_t off;  // if distance is further, it is off. 2 and 5
```
TomLonergan03, line 28: What are these for?

lib/sensors/brakes.hpp
```diff
@@ -0,0 +1,32 @@
...
+  ~Brakes();
+
+  //  bool open();
+  bool highLow();
+
+ private:
+  Brakes(core::ILogger &logger, std::shared_ptr<io::IGpioReader> gpio_reader);
+
+ private:
+  std::uint8_t pin_;
```
TomLonergan03, line 26: this can be `const`

TomLonergan03 (changes requested):

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+
+std::optional<bool> Brakes::isClamped()
+{
+  const auto optional_reading = gpio_reader_->read();
+  if (!optional_reading) { return std::nullopt; }
+  const auto reading = *optional_reading;
+
+  if (reading == core::DigitalSignal::kHigh) { return false; }
+  return true;
```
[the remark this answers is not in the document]
licornes-fluos: Nevermind

lib/sensors/brakes.cpp
```diff
@@ -0,0 +1,38 @@
...
+    return std::nullopt;
+  }
+  logger.log(core::LogLevel::kDebug, "Successfully created Brakes instance");
+  return Brakes(logger, *reader);
+}
+
+Brakes::Brakes(core::ILogger &logger, std::shared_ptr<io::IGpioReader> gpio_reader)
+    : gpio_reader_(gpio_reader),
+      logger_(logger),
+      pin_()
```
davidbeechey, line 21: Do you need this here? Either pass the pin into the constructor if it may need to be accessible in the future (i.e. if something else might want to know which pin it's using) or remove it from here and the header file

davidbeechey (changes requested):

PR #75, opened 2024-02-15
Infra - Pi Cross-Compiling with Docker

DylanCavers:
Add functionality to cross-compile for raspberry pi's to the Docker build system.

### DIFF
--- a/cc/Dockerfile.crosscompile
+++ b/cc/Dockerfile.crosscompile
@@ -1,11 +1,11 @@
-FROM ubuntu:latest
+FROM arm64v8/ubuntu:latest
 ARG DEBIAN_FRONTEND=noninteractive
 
 # Install dependencies
 RUN apt-get update && \
-    apt-get install -y git clang clang-format libboost-all-dev libeigen3-dev rapidjson-dev build-essential gcc make cmake libssl-dev libncurses5-dev libncursesw5-dev sudo
-    
-# Install Paho C++
+    apt-get install -y git clang libboost-all-dev libeigen3-dev rapidjson-dev build-essential gcc make cmake libssl-dev libncurses5-dev libncursesw5-dev sudo
+
+# Paho
 WORKDIR /home
 RUN git clone --recurse-submodules https://github.com/eclipse/paho.mqtt.cpp.git
 WORKDIR /home/paho.mqtt.cpp
@@ -15,6 +15,7 @@ RUN sudo cmake --build build/ --target install
 RUN sudo ldconfig
 
 WORKDIR /home/hyped_entrypoint
+
 COPY entry.sh ./
 
 ENTRYPOINT [ "/home/hyped_entrypoint/entry.sh" ]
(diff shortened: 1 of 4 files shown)


entry.sh
```diff
@@ -16,7 +14,15 @@ elif [ ! -d "build" ]; then
...
 
-cd build 
-cmake ..
-make -j
-make test
+if [[ $CROSS_COMPILE = true ]]; then
+    echo "Cross compiling for Raspberry Pi..."
+    cd build 
+    cmake  -DCMAKE_CXX_COMPILER=aarch64-rpi3-linux-gnu-g++ -DCURSES_LIBRARY=/usr/lib/x86_64-linux-gnu/libncurses.so -DCURSES_INCLUDE_PATH=/usr/include ..
+    make -j
```
TomLonergan03, line 21: cos this is an arm container, you can run `make test` also right?

DylanCavers: pretty sure we can, but may be overkill as we presumably won't need to run the tests at this stage and will be fairly computationally expensive cause of emulation. Unless we want it as proof the cross-compiled code works.

docker.sh
```diff
@@ -83,22 +76,62 @@ handle_options() {
...
+    docker buildx build -f cc/Dockerfile.crosscompile -t $CC_IMAGE_NAME .
+  else
+    echo "[!] Rebuild image"
+    docker build -t $IMAGE_NAME .
+  fi
+else
+  # Only build if the image does not exist
+  if [ "$cross_compile" = true ]; then
+    image=$( docker images -q $CC_IMAGE_NAME 2> /dev/null )
+    echo $image
```
davidbeechey, line 93: was this just for debugging?

cc/Dockerfile.crosscompile
davidbeechey, line 1: is there a reason this is in its own folder?

davidbeechey (approved):
Looks good :)

docker.sh
DylanCavers: ah yeah will remove

cc/Dockerfile.crosscompile
DylanCavers: initially for neatness as I had config files etc. But now its kinda pointless, can either clean up the other docker stuff into a Docker folder or move this dockerfile out to the root directory

entry.sh
```diff
@@ -18,5 +18,5 @@ fi
 
 cd build 
 cmake ..
-make
+make -j4
```
TomLonergan03, line 21: just `-j` or do some experimenting to find the fastest number based on `nprocs` (I've heard `nprocs`/`nprocs+1`/`nprocs*2` are all potentials)

TomLonergan03 (changes requested):

entry.sh
DylanCavers: cool will investigate, `-j4` was the highest I could go without it failing

TomLonergan03 (approved):
have u tested with a pi?

PR #81, opened 2024-02-26
HPRelay

emmabaird:
Referred to hardware.gpio and keyence which alsouses GPIO for 1st attempt to program to write to GPIO. There are a couple of cannot open source
file and #include errors I'm not sure how to
solve.

### DIFF
--- a/lib/sensors/HPRelay.cpp
+++ b/lib/sensors/HPRelay.cpp
@@ -1,0 +1,46 @@
+#include "HPRelay.hpp"
+
+namespace hyped::sensors {
+
+HPRelay::HPRelay(core::ILogger &log) : logger_(log)
+{
+}
+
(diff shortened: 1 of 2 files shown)


TomLonergan03 (changes requested):
good start! I think there's a nicer abstraction where the client interacts with a relay by simply doing
```c++
auto optional_relay = Sensors::Relay::create(/* args */);
// do the optional checks
auto relay = *optional_relay;
relay.close();
// do the high power stuff or whatever
relay.open();
```

licornes-fluos:
Change file names to `relay.hpp` and `relay.cpp`

licornes-fluos (changes requested):
I think you should start from scratch and look at [#72](https://github.com/Hyp-ed/hyped-2024/pull/72 ).
Your code is overcomplicated and somehow doesn't use the protocols properly. 
You need to call methods from the gpio protocol (hyped-2024/lib/io/gpio.hpp) and use those in your code.

PR #87, opened 2024-03-16
TEL - Fix end of line characters

davidbeechey:
Fixes `.prettierrc` file so that it sets end of line character to `lf` not `auto`.


TomLonergan03 (approved):
jfc

licornes-fluos (approved):

PR #93, opened 2024-04-02
Infra - Ban CRLF

TomLonergan03:
Decided to kill them for good.
Run `git config --global core.autocrlf input` to make git automatically convert line endings to LF when they are committed.


davidbeechey (approved):
With my sincere apologies,
David really-needs-to-switch-to-linux Beechey

DylanCavers (approved):

PR #97, opened 2024-04-05
SNS - Inverter Current

davidbeechey:
Sensors implementation for the inverter current sensor.

Datasheet: https://www.lem.com/sites/default/files/products_datasheets/hob-p_series.pdf

### DIFF
--- a/lib/sensors/inverter_current.cpp
+++ b/lib/sensors/inverter_current.cpp
@@ -1,0 +1,38 @@
+#include <inverter_current.hpp>
+#include "adc_mux.hpp"
+
+namespace hyped::sensors {
+
+std::optional<InverterCurrent> InverterCurrent::create(core::ILogger &logger,
+                                                       std::shared_ptr<io::II2c> i2c,
+                                                       const std::uint8_t adc_mux_channel)
+{
+  return InverterCurrent(logger, i2c, adc_mux_channel);
+}
+
+InverterCurrent::InverterCurrent(core::ILogger &logger,
+                                 std::shared_ptr<io::II2c> i2c,
+                                 const std::uint8_t adc_mux_channel)
+    : logger_(logger),
+      i2c_(i2c),
+      adc_mux_channel_(adc_mux_channel)
+{
+}
+
+InverterCurrent::~InverterCurrent()
+{
+}
+
+std::optional<core::Float> InverterCurrent::readCurrent()
+{
+  const auto inverter_current = i2c_->readByte(kAdcMuxAddress, adc_mux_channel_);
+  if (!inverter_current) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read inverter current");
+    return std::nullopt;
+  }
+  logger_.log(core::LogLevel::kDebug, "Successfully read inverter current");
+  // Output is 0-5V, which corresponds to 50-130A
+  return static_cast<core::Float>(*inverter_current) * 16 + 50;
+}
+
+}  // namespace hyped::sensors
--- a/lib/sensors/adc_mux.hpp
+++ b/lib/sensors/adc_mux.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cstdint>
+
 namespace hyped::sensors {
 
 constexpr std::uint8_t kAdcMuxAddress  = 0x1D;
--- a/lib/sensors/inverter_current.hpp
+++ b/lib/sensors/inverter_current.hpp
@@ -1,0 +1,33 @@
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+class InverterCurrent {
+ public:
+  static std::optional<InverterCurrent> create(core::ILogger &logger,
+                                               std::shared_ptr<io::II2c> i2c,
+                                               const std::uint8_t adc_mux_channel);
+  ~InverterCurrent();
+
+  std::optional<core::Float> readCurrent();
+
+ private:
+  InverterCurrent(core::ILogger &logger,
+                  std::shared_ptr<io::II2c> i2c,
+                  const std::uint8_t adc_mux_channel);
+
+ private:
+  core::ILogger &logger_;
+  std::shared_ptr<io::II2c> i2c_;
+  const std::uint8_t adc_mux_channel_;
+};
+
+}  // namespace hyped::sensors


lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,38 @@
+#include <inverter_current.hpp>
+#include "adc_mux.hpp"
+
+namespace hyped::sensors {
+
+std::optional<InverterCurrent> InverterCurrent::create(core::ILogger &logger,
```
TomLonergan03, line 7: you don't need a create method unless there is stuff that can cause the constructor to fail

lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,38 @@
...
+                                                       std::shared_ptr<io::II2c> i2c,
+                                                       const std::uint8_t adc_mux_channel)
```
TomLonergan03, line 9: an idea I just thought of here, change the `kAdcMuxChannelN` values to be an enum where each option has the value of the channel address. then here you have a type guarantee that it will be a valid register.

lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+
+std::optional<core::Float> InverterCurrent::readCurrent()
+{
+  const auto inverter_current = i2c_->readByte(kAdcMuxAddress, adc_mux_channel_);
+  if (!inverter_current) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read inverter current");
+    return std::nullopt;
+  }
+  logger_.log(core::LogLevel::kDebug, "Successfully read inverter current");
```
TomLonergan03, line 34: this log isn't necessary, it would be pretty spammy

TomLonergan03 (changes requested):
very good start, a few things that can be nicer (I should have thought a bit harder about the adc mux tbh)

lib/sensors/inverter_current.cpp
davidbeechey: I have removed it, but similar logs are in other sensors if you want me to also remove those?

TomLonergan03: Leave them for now but we're going to need to do a big cleanup at some point soon

TomLonergan03 (changes requested):
This is all good, one more thing to do is add a test command to the [debug REPL](https://github.com/Hyp-ed/hyped-2024/wiki/Debug-REPL). You need to get the i2c object, then construct the sensor, and create a command that reads from the sensor, as well as doing the TOML parsing required.

lib/debug/commands/inverter_current_commands.cpp
```diff
@@ -0,0 +1,54 @@
...
+  // Get the I2C instance
+  const auto optional_i2c = repl->getI2c(bus);
+  if (!optional_i2c) {
+    logger.log(core::LogLevel::kFatal, "Error getting I2C bus");
+    return core::Result::kFailure;
+  };
+  const auto i2c = std::move(*optional_i2c);
+
+  // Create the sensor instance
+  const auto sensor = std::make_shared<sensors::InverterCurrent>(
```
TomLonergan03, line 34: `current_sensor`

lib/debug/commands/inverter_current_commands.cpp
```diff
@@ -0,0 +1,54 @@
...
+    logger, i2c, static_cast<sensors::AdcMuxChannel>(adc_mux_channel));
+
+  // Create the read command
+  const auto read_command_name        = "read";
```
TomLonergan03, line 38: `"current read"` or something similar

TomLonergan03 (approved):
Looks good, just needs lab time

licornes-fluos (commented):
Other than the changes requested by Tom, all seems fine

lib/debug/commands/inverter_current_commands.cpp
licornes-fluos: Also should not be a constant

lib/debug/commands/inverter_current_commands.cpp
```diff
@@ -0,0 +1,54 @@
...
+  const auto read_command_handler     = [&logger, inverter_current_sensor]() {
+    const auto current = inverter_current_sensor->readCurrent();
+    if (!current) {
+      logger.log(core::LogLevel::kFatal, "Failed to read inverter current");
+      return;
+    }
+    logger.log(core::LogLevel::kDebug, "Inverter current: %d", *current);
+  };
+  auto read_command
+    = std::make_unique<Command>(read_command_name, read_command_description, read_command_handler);
```
TomLonergan03, line 49: There should also be a usage entry here

lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,54 @@
...
+  // The sensor maps -75A to 75A into -3V to 3V
+  // Since the ADC isn't differential, we need to differentiate the reference signal and the output
+  // signal
+  static const int MAX_CURRENT         = 75;
+  static const int MIN_CURRENT         = -75;
+  static const core::Float MAX_VOLTAGE = 3.3;
+  static const core::Float MIN_VOLTAGE = -3.3;
+
+  // These can be adjusted
+  const int reference_voltage      = 1.65;
```
TomLonergan03, line 40: `const int ... = 1.65;`

most type understanding webdev smdh (i'm totally not one now)

TomLonergan03, line 40: all of these should be in the header

lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,54 @@
...
+  const core::Float virtual_ground = MAX_VOLTAGE / 2;
+
+  // Calculate the current
+  const int inverter_current_voltage
+    = *inverter_current * (MAX_VOLTAGE - virtual_ground) + virtual_ground;
+  const core::Float current
+    = inverter_current_voltage * ((MAX_CURRENT - MIN_CURRENT) / (MAX_VOLTAGE - MIN_VOLTAGE))
+      + MIN_CURRENT;
+  const core::Float reference_voltage_biased = reference_voltage + virtual_ground;
+  const core::Float current_difference       = current - reference_voltage_biased;
```
TomLonergan03, line 50: what does this do? I doubt it's valid to subtract a voltage from an amperage

lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,54 @@
...
+    = i2c_->readByte(kAdcMuxAddress, static_cast<std::uint8_t>(adc_mux_channel_));
+  if (!inverter_current) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read inverter current");
+    return std::nullopt;
+  }
+
+  // The sensor maps -75A to 75A into -3V to 3V
+  // Since the ADC isn't differential, we need to differentiate the reference signal and the output
+  // signal
+  static const int MAX_CURRENT         = 75;
```
TomLonergan03, line 34: `int` should be sized, `std::uint32_t` or whatever suits

lib/sensors/inverter_current.cpp
```diff
@@ -0,0 +1,54 @@
...
+  static const int MIN_CURRENT         = -75;
+  static const core::Float MAX_VOLTAGE = 3.3;
+  static const core::Float MIN_VOLTAGE = -3.3;
+
+  // These can be adjusted
+  const int reference_voltage      = 1.65;
+  const core::Float virtual_ground = MAX_VOLTAGE / 2;
```
TomLonergan03, line 41: what is the difference between reference voltage and virtual ground?

lib/sensors/inverter_current.cpp
davidbeechey: 💀

davidbeechey: good point

PR #98, opened 2024-04-09
SNS - Accelerometer REPL Commands

licornes-fluos:
As we are using the same accelerometer code as last year, here is the code to test it with the current accelerometer.


lib/debug/commands/accelerometer_commands.cpp
```diff
@@ -0,0 +1,44 @@
...
+  const auto optional_i2c = repl->getI2c(bus);
+  if (!optional_i2c) {
+    logger.log(core::LogLevel::kFatal, "Error creating I2C bus");
+    return core::Result::kFailure;
+  };
+  const auto i2c = std::move(*optional_i2c);
+
+  // instance of sensor
+  const auto accelerometer_sensor           = std::make_shared<sensors::Accelerometer>(logger, i2c);
+  const auto accelerometer_read             = "read";
```
TomLonergan03, line 28: var name `accelerometer_read_name` and set the value `"accelerometer read"`

lib/debug/commands/accelerometer_commands.cpp
```diff
@@ -0,0 +1,44 @@
...
+  const auto accelerometer_read_description = "Read accelerometer";
+  const auto accelerometer_read_handler     = [&logger, accelerometer_sensor]() {
+    const auto value_ready  = accelerometer_sensor->isValueReady();  // method already returns state
```
TomLonergan03, line 31: this should be used, also what does the comment mean?

lib/debug/commands/accelerometer_commands.cpp
```diff
@@ -0,0 +1,44 @@
...
+    const auto acceleration = accelerometer_sensor->read();
+    if (!acceleration) {
+      logger.log(core::LogLevel::kFatal, "Failed to read acceleration");
+      return;
+    }
+    logger.log(core::LogLevel::kDebug, "Acceleration: %d", *acceleration);
```
TomLonergan03, line 37: `*acceleration` is a struct, i don't think you can print it like that

TomLonergan03 (changes requested):

lib/debug/commands/accelerometer_commands.cpp
```diff
@@ -0,0 +1,63 @@
...
+  auto accelerometer_sensor           = std::move(*optional_accelerometer_sensor);
+  const auto read_command_name        = "accelerometer read";
+  const auto read_command_description = "Read from the accelerometer";
+  const auto read_command_handler     = [&logger, accelerometer_sensor]() {
+    const auto value_ready = accelerometer_sensor.isValueReady();
+    if (!value_ready) {
+      logger.log(core::LogLevel::kFatal, "Value is not ready");
+      return;
+    }
+    const auto acceleration = accelerometer_sensor.read();
```
licornes-fluos, line 50: How to I call this? I tried multiple ways, but it doesn't work. (same for value_ready above)

lib/sensors/accelerometer.hpp
```diff
@@ -17,16 +17,15 @@ namespace hyped::sensors {
...
+enum class accelerometerAddress { k1D = 0x1D, k1E = 0x1E };
 
 class Accelerometer : public II2cMuxSensor<core::RawAccelerationData> {
  public:
   static std::optional<Accelerometer> create(core::ILogger &logger,
                                              std::shared_ptr<io::II2c> i2c,
                                              const std::uint8_t channel,
-                                             const std::uint8_t device_address);
+                                             const accelerometerAddress device_address);
+  //  const std::uint8_t device_address);
```
davidbeechey, line 28:
```suggestion
```

davidbeechey (approved):

davidbeechey (approved):

TomLonergan03 (approved):
lets gamble try merge

PR #99, opened 2024-04-09
SNS - Temperature sensor REPL commands

TomLonergan03:
This works!


davidbeechey (approved):

lib/sensors/temperature.hpp
```diff
@@ -11,16 +11,14 @@
 
 namespace hyped::sensors {
 
-// Two possible addresses for the temperature sensor
-constexpr std::uint8_t kDefaultTemperatureAddress     = 0x38;
-constexpr std::uint8_t kAlternativeTemperatureAddress = 0x3F;
+enum class temperatureAddress { k7F = 0x7F, k38 = 0x38, k3C = 0x3C, k3E = 0x3E };
```
licornes-fluos, line 14: According to the datasheet, the addresses are 0x70/78/7C/7E

licornes-fluos (changes requested):

lib/debug/commands/temperature_commands.cpp
```diff
@@ -0,0 +1,52 @@
...
+  };
+  const auto i2c              = std::move(*optional_i2c);
+  const auto optional_address = config["address"].value<std::uint8_t>();
+  if (!optional_address) {
+    logger.log(core::LogLevel::kFatal, "Invalid address");
+    return core::Result::kFailure;
+  }
+  const auto address                     = std::move(*optional_address);
+  const auto optional_temperature_sensor = sensors::Temperature::create(
+    logger, i2c, 0, static_cast<sensors::temperatureAddress>(address));
```
licornes-fluos, line 30: Shouldn't this be bus 2, as defined in the debug.toml?

lib/sensors/temperature.hpp
TomLonergan03: uh where are you seeing that? the sensor definitely shows up at 0x38

lib/debug/commands/temperature_commands.cpp
TomLonergan03: that's the mux channel. we'll merge this, then when the accelerometer PR gets merged there will be a cleanup

licornes-fluos (approved):

PR #109, opened 2024-04-18
Infra - Change from rapidjson to reflectcpp

TomLonergan03:
Rapidjson is a pain and overly boilerplatey, [reflect-cpp](https://github.com/getml/reflect-cpp) should be much nicer to use


TomLonergan03:
Giving up on this cos it's not really important

PR #112, opened 2024-04-29
Infra - Add a static analyser

TomLonergan03:
We finally be following the NASA rules (kinda)

Lotta issues found, gotta work out how strict we want to be and what the appropriate rules are, and fix all actual issues.

(Yes I have an exam in an hour, totally got nothing better to do)

- [x] Add clang-tidy
    - [x] Add .clang-tidy
    - [x] Fix errors
- [x] ~~Add include-what-you-use~~ seems more of a pain than it's worth ngl

# Build times
Built on my laptop with `./docker.sh -b`, a cached docker image, and an empty `./build/` folder

| Before | After |
|:-:|:-:|
| 40.45s | 188.50s |

It'll run on github as part of CI but will be run locally if cmake is run as `cmake .. -DLINT=ON`, or `./docker.sh -b -l`


lib/sensors/accelerometer.hpp
```diff
@@ -24,10 +24,10 @@ constexpr std::uint8_t kAlternativeAccelerometerAddress = 0x18;
 class Accelerometer : public II2cMuxSensor<core::RawAccelerationData> {
  public:
   static std::optional<Accelerometer> create(core::ILogger &logger,
-                                             std::shared_ptr<io::II2c> i2c,
+                                             const std::shared_ptr<io::II2c> &i2c,
                                              const std::uint8_t channel,
                                              const std::uint8_t device_address);
-  ~Accelerometer();
+  ~Accelerometer() override = default;
```
licornes-fluos, line 30: How does this work and why isn't it in the cpp file anymore?

licornes-fluos (approved):
I'll need explanations on a few things that changed to see how it affects all the code that isn't merged.

lib/sensors/accelerometer.hpp
TomLonergan03: default destructor is the same as giving it a body that is just `{}`. I moved it to the header cos it might as well be there tbh, not much point in having an `~Accelerometer() = default` entry in the cpp

I should check what the difference between declaring it at all or not is

TomLonergan03: declaring things `default` does nothing unless we're overloading constructors - which we aren't. I've removed them all except in the i2cmux template where it is needed to compile

DylanCavers (approved):

lsalhani (approved):
all good

misha7b (approved):
Nav files look good

PR #119, opened 2024-05-24
NAV-MQTT

misha7b:
Created main run loop for NAV.
Dummy variables currently used, not reading the payload yet.


lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+  auto message_payload = std::make_shared<rapidjson::Document>();
+  message_payload->SetObject();
+
+  auto trajectory = currentTrajectory();
+  if (!trajectory) { logger_.log(core::LogLevel::kFatal, "Failed to get current trajectory"); }
+
+  rapidjson::Value displacement(trajectory->displacement);
+  rapidjson::Value velocity(trajectory->velocity);
+  message_payload->AddMember("displacement", displacement, message_payload->GetAllocator());
+  message_payload->AddMember("velocity", velocity, message_payload->GetAllocator());
```
TomLonergan03, line 147: will there be an acceleration value as well?

lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+  mqtt_->subscribe(core::MqttTopic::kKeyence);
+  mqtt_->subscribe(core::MqttTopic::kOpticalFlow);
+  mqtt_->subscribe(core::MqttTopic::kAccelerometer);
+
+  while (true) {
+    mqtt_->consume();
+
+    auto keyence_msg       = mqtt_->getMessage();
+    auto optical_msg       = mqtt_->getMessage();
+    auto accelerometer_msg = mqtt_->getMessage();
```
TomLonergan03, line 166: this won't work, each time you call `getMessage()` you get whatever the next message is. you then need to look at what topic that message is from, then do something with it. I think you'll want to store the value until you have a fresh acceleration and a fresh optical velocity measurement then run the filter over it once.

Easiest way I can think of is have `Navigator` have a `std::optional<std::uint16_t> most_recent_keyence_` etc, then you set each to a value when the message comes in, and if at least `most_recent_accerometer_` and `most_recent_optical_flow_` have values then run the filter and set them back to `std::nullopt` afterwards

lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+      = {core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false),
+         core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false),
+         core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false),
+         core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false)};
+
+    if (keyenceUpdate(dummy_keyence_data) == core::Result::kFailure
+        || opticalUpdate(dummy_optical_data) == core::Result::kFailure
+        || accelerometerUpdate(dummy_accelerometer_data) == core::Result::kFailure) {
+      logger_.log(core::LogLevel::kFatal, "Failed to update sensor data");
+      break;
```
TomLonergan03, line 189: send a message to the state machine topic requesting a failure if this happens

lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+
+    std::array<core::RawAccelerationData, core::kNumAccelerometers> dummy_accelerometer_data
+      = {core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false),
+         core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false),
+         core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false),
+         core::RawAccelerationData(0, 0, 0, core::TimePoint{}, false)};
+
+    if (keyenceUpdate(dummy_keyence_data) == core::Result::kFailure
+        || opticalUpdate(dummy_optical_data) == core::Result::kFailure
+        || accelerometerUpdate(dummy_accelerometer_data) == core::Result::kFailure) {
```
TomLonergan03, line 187: do `keyence_update_result = keyenceUpdate()` etc, then the if statement instead of blocking it all together

lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+      logger_.log(core::LogLevel::kFatal, "Failed to update sensor data");
+      break;
+    }
+  }
+
+  publishCurrentTrajectory();
```
TomLonergan03, line 193: this is outside the while loop, pretty sure it should be inside it

lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+                                         .priority  = core::MqttMessagePriority::kNormal};
+  const core::MqttMessage message{topic, header, message_payload};
+  mqtt_->publish(message, core::MqttMessageQos::kExactlyOnce);
+}
+
+void Navigator::run()
+{
+  mqtt_->subscribe(core::MqttTopic::kKeyence);
+  mqtt_->subscribe(core::MqttTopic::kOpticalFlow);
+  mqtt_->subscribe(core::MqttTopic::kAccelerometer);
```
TomLonergan03, line 159: these can all fail, and should return if any of them fail

lib/navigation/control/navigator.cpp
```diff
@@ -127,4 +132,65 @@ core::Result Navigator::accelerometerUpdate(
...
+
```
TomLonergan03, line 160: can you make a `kStarted` topic and send a message on it saying that the navigation module has started successfully, we need to add something to deal with that stuff somewhere

TomLonergan03 (changes requested):
another thing to do is add an entry in the config file, and edit `src/pod/main.cpp` to spin up that node if the current host should have a navigator node. look at the state machine one for inspiration

TomLonergan03 (approved):
LGTM (not a self approve lmao)

PR #121, opened 2024-05-25
TEL - Environment variable handling

davidbeechey:
An attempt to improve how we handle environment variables (to include validation, build-time checking and easier use with Docker).

To do:
- [ ] Fix runtime env use - will probably need to split up env schemas for different packages so they have access to runtime env


davidbeechey:
Going to close this for this year - it's something we could look at again next year. Going to be more complex than it's worth to get working in a way that's significantly more useful than what we have at the moment.

PR #127, opened 2024-07-02
TEL - Create Latency Node

davidbeechey:


telemetry/packages/ui/app/context/pods.tsx
```diff
@@ -152,7 +152,7 @@ export const PodsProvider = ({ children }: { children: React.ReactNode }) => {
           publish(
             'latency/request',
             JSON.stringify({
-              latency: new Date().getTime().toString(),
+              timestamp: new Date().getTime().toString(),
```
TomLonergan03, line 155: here and elsewhere, are you remembering that all mqtt messages are of the format:
```json
{
  header : {
    timestamp: int,
    priority: int
  },
  payload: { rest of data }
}
```

TomLonergan03 (changes requested):

telemetry/packages/ui/app/context/pods.tsx
davidbeechey: Ahhh I see. Will need to change this throughout the telemetry code

